Issue #215, opened 2013-04-30
Improve throughput by reducing number of calls to stats_half_start/stats_half_end


justfalter:
Right now, stats_half_start and stats_half_end are called for each packet. In my testing, this accounts for ~13% of ettercap's processing time. If we were to sample once every 100 packets rather than once per packet, we'd significantly reduce it's impact upon Ettercap's performance.

koeppea:
@justfalter: Can you check #449 which request to merge the proposed change.
I've done a test case. Despite the bad throughput performance of Ettercap in general, it seems your promising idea didn't work out.
Could you please have a look. Maybe I've done something wrong.
Thanks.

PR #449, opened 2014-01-07
Fixing issue #215

LocutusOfBorg:
I don't know, does this code makes sense?

I would like to test it, but I don't know how to do it properly

### DIFF
--- a/include/ec_globals.h
+++ b/include/ec_globals.h
@@ -46,6 +46,7 @@ struct ec_conf {
    int connection_buffer;
    int connect_timeout;
    int sampling_rate;
+   int packet_update_count;
    int close_on_eof;
    int aggressive_dissectors;
    int skip_forwarded;
--- a/share/etter.conf.v4
+++ b/share/etter.conf.v4
@@ -37,6 +37,7 @@ connect_timeout = 5           # seconds
 
 [stats]
 sampling_rate = 50            # number of packets 
+packet_update_count = 50      # number of packets
 
 [misc]
 close_on_eof = 1              # boolean value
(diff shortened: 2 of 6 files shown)


koeppea:
Rebased onto current master and renamed commit.
Need to test if it really has a positive effect when ettercap has to serve some bandwidth.

koeppea:
Rebased over latest master and resolved conflicts + two errors.
Testing now the impact on transfer performance...

koeppea:
Comparism test done:
```
1GByte via FTP from one VM(FreeBSD) to another VM(Debian 64-bit) on the same host
w/o Ettercap(IPv6):         duration: 00:50 (19.68 MiB/s)
w/o Ettercap(IPv4):         duration: 00:46 (21.63 MiB/s)
w/  Ettercap(master/IPv6):  duration: 04:45 ( 3.50 MiB/s)
w/  Ettercap(master/IPv4):  duration: 04:25 ( 3.76 MiB/s)
w/  Ettercap(fix-215/IPv4): duration: 04:44 ( 3.52 MiB/s)
w/  Ettercap(fix-215/IPv6): duration: 04:31 ( 3.68 MiB/s)
```
I also promised myself more out of it. However, since it's improving in both cases (IPv4 & IPv6) against master, I vote for merge in any case.

sgeto:
> IPv4 even worse.

Worse than before applying this PR?

koeppea:
Well I think this PR may have an effect on performance.
But at the moment the issue lies somewhere else. I'm afraid now it's time to profile Ettercap to find the real bottleneck. When this is found and potentially fixed, this PR could make a difference, then relativiately to the improved throughput.

PR #522, opened 2014-04-09
Fixing Lack of Sanity checks for malloc() calls

LocutusOfBorg:


gvanem:
I have an itch with the use of some Gtk memory. In *interfaces/gtk/ec_gtk_conf.c*, the allocated `filename`:
```
   filename = g_build_filename(path, "ettercap_gtk", NULL);
```
should be freed by Gtk. Hence:
```
@@ -103,7 +103,7 @@
       fclose(fd);
    }

-   free(filename);
+   g_free(filename);
    filename = NULL;
 }
```
Prior to this patch, it caused a crash here since I mixed debug/release mode in Ettercap.

(I add this as a comment here instead of openng a new issue).

dogbert2:
That makes sense, g_free() is the better option, rather than free() since
it is in GTK (my bad)...

LocutusOfBorg:
applied thanks!

LocutusOfBorg:
#677

PR #544, opened 2014-05-02
moving lua to ettercap directory

LocutusOfBorg:
@justfalter what do you think about this?

having the lua stuff in /usr/share/lua instead of /usr/share/ettercap/lua for me is a BAD idea...

Can you please test and give feedbaks?

### DIFF
--- a/include/ec_lua.h
+++ b/include/ec_lua.h
@@ -8,6 +8,8 @@ EC_API_EXTERN int ec_lua_init();
 EC_API_EXTERN int ec_lua_fini();
 EC_API_EXTERN int ec_lua_cli_add_script(char * script);
 EC_API_EXTERN int ec_lua_cli_add_args(char * args);
+EC_API_EXTERN void ec_lua_print_info(FILE* debug_file);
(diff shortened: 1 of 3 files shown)


gvanem:
Sorry to be a pain in the butt. But the path-mess on Windows still confuses me (not sure if it's simpler oon *unix). On Windows, a command like:
```
f:\MingW32\src\inet\Ettercap\src>..\ettercap.exe -v
```
gives:
```
...
-> ${INSTALL_LUA_ROOT}    share/lua/
-> ${INSTALL_LUA_INIT}    share/lua/init.lua
-> ${INSTALL_LUA_CORE}    share/lua/
-> ${INSTALL_LUA_SCRIPTS} share/lua/
-> ettercap 0.8.2
...
```

AFAICS, only `INSTALL_LUA_INIT` is needed at run-time. So shouldn't this use `get_local_path ("lua/init.lua")` instead?

LocutusOfBorg:
@gvanem can you please share a patch? I think I agree there

PR #635, opened 2014-12-18
Followup for pull 607

LocutusOfBorg:


LocutusOfBorg:
@koeppea what do you think?

koeppea:
Looks good to me.

PR #646, opened 2015-01-10
Freebsd buildfix

koeppea:
This pull request adjusts a number of issues to build ettercap on FreeBSD.
This includes as well the update on the specific platform documentation.


src/os/ec_bsd.c
```diff
@@ -136,7 +136,7 @@ u_int16 get_iface_mtu(const char *iface)
 
    /* open the socket to work on */
    sock = socket(PF_INET, SOCK_DGRAM, 0);
-   if (sock = -1)
+   if (sock == -1)
```
LocutusOfBorg: @koeppea basically you are saying that our BSD port never worked? :)
or at least the get mtu never did ;)

koeppea: I don't know how the FreeBSD guys build  the package. I learned, that they take the existing source and patch from upstream. However this would normally also end in build issues.
However I started to test ettercap on FreeBSD and faced all the different addressed issues.

This specific issue was just a warning. However the actual functionality was obviously not given.

cmake/Modules/EttercapLibCheck.cmake
```diff
@@ -122,7 +122,8 @@ if(ENABLE_PLUGINS)
 #        set(HAVE_PLUGINS 1)
 #        set(EC_LIBS ${EC_LIBS} ${HAVE_LTDL})
 #    endif(HAVE_LTDL)
-    CHECK_LIBRARY_EXISTS(dl dlopen "dlfcn.h" HAVE_DLOPEN)
+    #CHECK_LIBRARY_EXISTS(dl dlopen "dlfcn.h" HAVE_DLOPEN)
+    CHECK_FUNCTION_EXISTS(dlopen HAVE_DLOPEN)
```
LocutusOfBorg: mmmm I found probably this CMAKE_DL_LIBS that might do all for us
http://www.cmake.org/cmake/help/v3.0/variable/CMAKE_DL_LIBS.html

what do you think about?

koeppea: let me test this....

cmake/Modules/EttercapLibCheck.cmake
```diff
@@ -117,22 +117,18 @@ endif(HAVE_ICONV)
 
 # LTDL
```
koeppea: @LocutusOfBorg how about that?

LocutusOfBorg: I would rather say something like:
```
if(ENABLE_PLUGINS)
  set(HAVE_PLUGINS 1)
  set(EC_LIBS ${EC_LIBS} ${CMAKE_DL_LIBS})
  set(EC_LIBETTERCAP_LIBS ${EC_LIBETTERCAP_LIBS} ${CMAKE_DL_LIBS})
endif(ENABLE_PLUGINS)
```

the variables are empty if not found, and I guess cmake is able to find CMAKE_DL_LIBS also in dlopen.

So I guess everytime is found and goes in the first branch, hence we can drop it :)
what is your opinion?

koeppea: How about systems where libdl is not installed? I guess this would result in a build failure. This was where I was aiming at. 

Or we have to make the availability of libdl mandatory.

LocutusOfBorg: you are right, so:
```
if(ENABLE_PLUGINS)
  if(CMAKE_DL_LIBS)
    set(HAVE_PLUGINS 1)
    set(EC_LIBS ${EC_LIBS} ${CMAKE_DL_LIBS})
    set(EC_LIBETTERCAP_LIBS ${EC_LIBETTERCAP_LIBS} ${CMAKE_DL_LIBS})
  else(CMAKE_DL_LIBS)
     set(HAVE_PLUGINS 0)
     message("You requested to enable plugins, but dlopen function can't be found")
  endif(CMAKE_DL_LIBS)
endif(ENABLE_PLUGINS)

koeppea: On FeeBSD for example this variable would be empty because dlopen is part of libc. Your latest approach would disable plugin support on FreeBSD at least.

LocutusOfBorg: Ok thanks for the clarification, so :100: for your change :)

koeppea:
tested on Debian Jessie Linux as well on FreeBSD 10.1.
Ready to go from my side.

LocutusOfBorg:
Merged ^2, thanks another TON! :D

PR #647, opened 2015-01-10
Freebsd pfctl

koeppea:
This pull request fixes the handling of the Packet Filter of FreeBSD.
The way, `pfctl` is being used, requires more functionality on the redirect*_command options in *etter.conf*.
To make this kind of automatic / transparent handling work, the execution of the redirect*_command need to be enabled for piping support.
Furthermore, by accident I observed a wrong port number for FTPS being hard coded in the FTP dissector.


koeppea:
tested on FreeBSD 10.1 (in combination with #646) using the sslstrip plugin.
Ready to go from my side.

LocutusOfBorg:
merged, thanks a TON :)

PR #652, opened 2015-02-01
Fix possible flag full name error

new23d:
Else we get:

ettercap: unrecognized option '--nopromisc'


koeppea:
@new23d Thanks. Good catch. :+1:

LocutusOfBorg:
:+1:

LocutusOfBorg:
Really a nice catch! Can't wait for merge :)

PR #654, opened 2015-02-04
Reset logfile ownership upon dropping privileges

mmaxs:
If logging is started from the GUI menu item the log files are created after privileges have been dropped by the ettercap and the user with dropped UID:GID becomes the owner of the created log files.
In contrast, the log files specified on the command line remain to be owned by the root user even if privileges are dropped.
This proposal is about to reset the ownership of the log files each time when dropping the privileges to the new UID:GID.

### DIFF
--- a/src/ec_log.c
+++ b/src/ec_log.c
@@ -37,8 +37,11 @@
 
 /* globals */
 
-static struct log_fd fdp;
-static struct log_fd fdi;
+/* zero is formally a valid value for an opened file descriptor
+ * so we need a custom initializer
+ */
+static struct log_fd fdp = {0, NULL, -1};
+static struct log_fd fdi = {0, NULL, -1};
 
 /* protos */
 
@@ -79,7 +82,7 @@ int set_loglevel(int level, char *filename)
    /* all the host type will be unknown, warn the user */
    if (GBL_OPTIONS->read) {
       USER_MSG("*********************************************************\n");
-      USER_MSG("WARNING: while reading form file we cannot determine    \n");
+      USER_MSG("WARNING: while reading form file we cannot determine     \n");
       USER_MSG("if an host is local or not because the ip address of     \n");
       USER_MSG("the NIC may have been changed from the time of the dump. \n");
       USER_MSG("*********************************************************\n\n");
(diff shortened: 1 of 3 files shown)


src/ec_log.c
```diff
@@ -37,8 +37,11 @@
 
 /* globals */
 
-static struct log_fd fdp;
-static struct log_fd fdi;
+/* zero is formally a valid value for an opened file descriptor
```
koeppea: I thought 0 is reserved for stdin in terms of file descriptors?

koeppea:
Without having this tested yet, I like this pull request. Good job.

src/ec_log.c
mmaxs: Yep, as it is well known, it is preconnected to stdin. But a user can close this descriptor at the program launching, e.g. from the bash command line
$ ettercap 0<&-
And posix standard  says that: "The open() function shall return a file descriptor for the named file that is the lowest file descriptor not currently open for that process."
I don't know how it would be in real life -- whether open() will return 0 in the case if it is a lowest free file descriptor or not.

koeppea:
When compiling I get some minor warnings:
```
/home/koeppea/dev/ettercap/src/ec_log.c: In function ‘reset_logfile_owners’:
/home/koeppea/dev/ettercap/src/ec_log.c:245:48: warning: signed and unsigned type in conditional expression [-Wsign-compare]
/home/koeppea/dev/ettercap/src/ec_log.c:246:48: warning: signed and unsigned type in conditional expression [-Wsign-compare]
/home/koeppea/dev/ettercap/src/ec_log.c:260:48: warning: signed and unsigned type in conditional expression [-Wsign-compare]
/home/koeppea/dev/ettercap/src/ec_log.c:261:48: warning: signed and unsigned type in conditional expression [-Wsign-compare]
```
Could you please fix them?

koeppea:
The file permissions are ATM set to **0600**. Do you think it make sense to make it readable for every one: **0644**?

mmaxs:
In regard to this, shouldn't we rely on the umask of the process?
In other words, to create/open the log files with the mode:
S_IRUSR|S_IWUSR|S_IRGRP|S_IWGRP|S_IROTH|S_IWOTH
which will be automatically adjusted by user's umask setting as the he/she wants to.

LocutusOfBorg:
@koeppea what is your opinion on this one?

koeppea:
I fully support the **umask** approach. Setting the umask, perfectly controls the file permissions. So :+1: from me.

eaescob:
:+1:

PR #664, opened 2015-03-04
Fix OSX warning

LocutusOfBorg:


koeppea:
How is the `struct timeval` defined on a MacOSX?

LocutusOfBorg:
I don't know, maybe this way
http://macosx.com/threads/timeval-from-sys-resource-h-not-defined.12114/

anyway seems that also linux defines it in the same way, so I unified everything, lets wait for travis

koeppea:
Ah now I see where it's coming from... :smile:

LocutusOfBorg:
Nope, seems to be not working
https://travis-ci.org/Ettercap/ettercap/jobs/53163318

they are different, rebased and updated, lets see travis

koeppea:
Rebased over current Master

koeppea:
Ups. race condition

koeppea:
Hui. Wasn't finished.
Well nevermind. I'll create another pull request for that.

LocutusOfBorg:
yep, I prefer to merge because the delta with my fix-library branch becomes painful to maintain :)

PR #668, opened 2015-03-15
Fix overlink

LocutusOfBorg:
This pull request aims to remove some overlinking we created with libettercap.

In my opinion ettercap should link graphical stuff and libettercap.

while libettercap should link the core code and related libraries.


LocutusOfBorg:
lets merge :) we have already started this release cycle, so so much tests will follow

Issue #666, opened 2015-03-06
not enough resources to create a new thread in this process


daspri:
I do not have this problem with ettercap 0.8.0 or earlier. 

In 0.8.1 or git version the ncurses gui just stopped working after Viewing Connections for a while so I tried with "ettercap -T" after a while I get this error and it stops:

ERROR : 11, Resource temporarily unavailable
[ettercap/src/ec_threads.c:ec_thread_new_detached:210]
 not enough resources to create a new thread in this process: Operation not permitted

And in "top" the VIRT mem is 3gigs (which isn't he case in 0.8.0).

I am guessing it relates to the passive resolver (resolve hostnames is not enabled)

After that error I press CTRL-C and read the log, there are 760 "[ ettercap] ec_thread_kill_all -- terminating 2536287120 [resolv_passive ..." lines.

grep -i "terminating" ettercap-0.8.1_debug.log | wc -l
760

This is a 32bit system,  2.6.27-7

LocutusOfBorg:
can you please try with the latest git?

daspri:
I am using latest git. Same error in git and 0.8.1.

koeppea:
@daspri Could you please upload the debug logfile somewhere?

koeppea:
BTW: I'm in IRC ATM

koeppea:
Mhh I've ran ettercap in curses mode, poisening another host even with active name resolution enabled. I couldn't recognize an increase of memory usage for more that 20 minutes operation.
I've tested this on a Debian Jessie 32-bit VM.

Regarding the messages in the debug log: They're normal. When ettercap is quit, it kills all threads weather they've already been destroyed or not.

Which Distribution / version have you experienced this behaviour? If possible, do you also experience this behaviour on any other host?

daspri:
This is an old ubuntu box. 
Sometimes it happens in first few minutes, sometimes takes much longer. I don't know why yet.
Compiled without GTK/plugins/ipv6. Just curses in Debug mode.
Cleaned log (of IPs/hostnames) is here: http://pastebin.de/raw/?id=58615

koeppea:
Thanks for the debug log.
I saw in the logfile that you've called ettercap like `ettercap -T -u -r c2`. Is that correct?
So "*c2*" is a packet capture file? is it possible to get it after you put it through a pcap anonymizer?

koeppea:
As per our discussion on IRC:
I understand that the PCAP can not easily handed out.
[...]

PR #669, opened 2015-03-17
Check duplicate passive name recognition

koeppea:
This pull request intends to fix issue #666.
However it just addresses a symptom but the original issue is still there.

The name resolution should only be done only by a couple of threads which are reused for multiple queries. But this is still to be done as it's not such a quick win.

### DIFF
--- a/TODO
+++ b/TODO
@@ -54,6 +54,12 @@
       - decryption with given keys
       - DES support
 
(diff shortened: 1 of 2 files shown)


koeppea:
Closing as it does not solve the issue sufficiently.

PR #670, opened 2015-03-19
simplify passive name resolution

koeppea:
This pull requests eliminates the creation of a dedicated thread for each IP address seen in DNS replies. The original intention of this approach was to coordinate the cache writing of the active name resolution threads with the passive once.

The passive name resolution is only triggered by background threads (*capture*), hence the extra thread just to insert the ip-to-name combination into the cache isn't necessary. The only important thing is that the capture thread coordinates with the resolve_dns thread using the same mutex lock.

While this pull request addresses the passive name resolution, I already have a idea how to rework the active name resolution being more smart.

This pull requests mainly addresses the issue #666.

### DIFF
--- a/src/ec_resolv.c
+++ b/src/ec_resolv.c
@@ -48,7 +48,6 @@ struct resolv_entry {
 
 /* protos */
 EC_THREAD_FUNC(resolv_dns);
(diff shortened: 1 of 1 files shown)


koeppea:
Closing in favor of #672

PR #673, opened 2015-04-22
Update curl to 7.42

LocutusOfBorg:


koeppea:
Tested also with the fixed *sslstrip* code from #674. Works.

PR #678, opened 2015-05-01
Geoip support

LocutusOfBorg:


LocutusOfBorg:
still missing:
*add libgeoip-dev
*documentation
* how to handle the database? (geoip-database?)

we have a build failure related:
error: ‘struct ec_conf’ has no member named ‘geoip’
   if (!GBL_CONF->geoip.data_file)

LocutusOfBorg:
yeah, thanks! I'll apply it in a few hours,
src/ec_plugins have the MAX_PATH defined in a WINDOWS only code :)
"#ifdef OS_WINDOWS"

LocutusOfBorg:
I guess we should be almost ready some issues still to be addressed:
* use a data file found from cmake (in Debian there are some geoip-data with the db)
* fix the SET_POINTER
* Update also the curses interface.

the code received by mail should  be all committed, thanks again dear @gvanem for your precious contribution!

LocutusOfBorg:
Hi @gvanem can you please try
https://github.com/LocutusOfBorg/ettercap/commit/4592e675ac60c6d5e3e44bbd7e7d1fc10868f277

I guess GeoIP_new should take care of the file location (if installed in the standard way)

gvanem:
@LocutusOfBorg  (if installed in the standard way)

And what's the standard way on Windows? None.

LocutusOfBorg:
so if there is no standard way on windows, I presume GeoIP_new will fail and your code will be executed, can you please confirm that?
BTW when you build libgeoip if you pass the -DGEOIPDATADIR=\"/usr/share/GeoIP\" with the correct value, this is stored in the library, and used at runtime.

Why windows behaves differently? do you have a library build log?

koeppea:
Does cmake detection for libgeoip work for you?
When I run cmake, I only get *HAVE_GEOIP-NOTFOUND*. But I have it installed the standard way:
```
ii  geoip-database                       20150317-1                          all          IP lookup command line tools that use the GeoIP library (country database)
ii  libgeoip1:i386                       1.6.2-4                             i386         non-DNS IP-to-country resolver library
```
The database is located at: `/usr/share/GeoIP/GeoIP.dat`.
My cmake version is 3.0.2 (just upgraded to Jessie as I suspected the cmake version being responsible).

gvanem:
@LocutusOfBorg Why windows behaves differently? do you have a library build log?

No log. But hard-coded paths like **/usr/share/GeoIP** is just asking for trouble (even on Unix). First, since build-time and run-time directories can differ. We have drive-letters on Windows, remember. I've not tried your diff, but feel the `geoip_data_file`should be part of `etter.conf`. Then there would be no doubt where `Geoip.dat`is located.

Look at `geoip.c` and you'll notice the use of `GetModuleFileNameA()`; a heuristic hack to find the path relative to the application using GeoIP.

LocutusOfBorg:
>No log. But hard-coded paths like /usr/share/GeoIP is just asking for trouble (even on Unix). First, since >build-time and run-time directories can differ. We have drive-letters on Windows, remember. I've not
>tried your diff, but feel the geoip_data_fileshould be part of etter.conf. Then there would be no doubt >where Geoip.datis located.

you are completely right, but please look at the code:
I first try the "standard" way, and if I fail to find it, I use the method you describe as a fallback.

Is that ok for you? I presume normal users don't care about that file, so at least on unix they shouldn't be bothered of changing paths in the configuration file, or even updating the database

>Look at geoip.c and you'll notice the use of GetModuleFileNameA(); a heuristic hack to find the path >relative to the application using GeoIP.

yes, but this is a windows specific code :)
```
#if !defined(_WIN32)
        memset(path, 0, sizeof(char) * 1024);
        snprintf(path, sizeof(char) * 1024 - 1, "%s/%s", GEOIPDATADIR,
                 file_name);
#else
```

anyway, it is ok for me to keep the configuration value, just I would like to prior try to find the db in the library defined location.

LocutusOfBorg:
(I also added a better cleanup code, hope you agree)

LocutusOfBorg:
@koeppea 
>Does cmake detection for libgeoip work for you?
>When I run cmake, I only get HAVE_GEOIP-NOTFOUND. But I have it installed the standard way:
>ii  geoip-database                       20150317-1                          all          IP lookup command line tools that >use the GeoIP library (country database)
>ii  libgeoip1:i386                       1.6.2-4                             i386         non-DNS IP-to-country resolver library

geoip-database is needed at *runtime* not build time :)

you need libgeoip-dev in order to find the development files ;)

>My cmake version is 3.0.2 (just upgraded to Jessie as I suspected the cmake version being responsible).

BTW there is a 3.2.2 on unstable from yesterday, fixing a serious bug in findssl code I reported recently
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=784443

:)

koeppea:
I see there is some uncertainty regarding the configuration values. At the moment the configuration file can not be loaded properly.
May I suggest the following:
 1. Under *[misc]* add **geoip_enable** as a boolean value
 2. Add to the comment in the configuration file that the path to the database maybe adjusted in case the database is located in a non-standard location
 3. Add **geoip_data_file** under *[strings]*
 4. Add **geoip_data_file_v6** under *[strings]*

Of course this has to be changed in the current code. I can take this on.
IMHO this fits much better into the current configuration concept of ettercap.

What do you think?

LocutusOfBorg:
@koeppea I fully agree, the ipv6 stuff is something we should arrange, but I can't test it...

koeppea:
I can take this on as well. 
So will add a commit to align the configuration parameters first.

src/ec_conntrack.c
```diff
@@ -264,6 +265,18 @@ static void conntrack_add(struct packet_object *po)
    memcpy(&cl->co->L3_addr1, &po->L3.src, sizeof(struct ip_addr));
    memcpy(&cl->co->L3_addr2, &po->L3.dst, sizeof(struct ip_addr));
 
+   /* fill the 2 letter country-codes */
```
koeppea: I'd like to know your opinion: I'd prefer query the country at view time rather than at conntrack-time.
This way it saves time of the *top_half* thread which is the second most important thread in ettercap.
So I'd prefer calling *ec_get_country_from_ip()* within the user interface thread. Only at places where it is used.

koeppea: So something like calling *conntrack_countrystr()*.

LocutusOfBorg: reducing calls to the top_half is for sure something we should take care of :)

koeppea:
I added a few more things around this feature, fixed a integer underflow issue and rebased onto master.
In the next step I would begin with squashing some commits to get a meaningful commit history.
After that, I will check what's necessary to provide GeoIP support to etterlog as well.

LocutusOfBorg:
It works flawlessly for me! :+1:

LocutusOfBorg:
tested curses and gtk and text mode!

koeppea:
Alright. I think this commit history is squashed enough.

koeppea:
Tomorrow I'll check what's necessary for *etterlog*.

LocutusOfBorg:
@koeppea your merge wasn't working correctly... I hope I fixed your rebase :)

LocutusOfBorg:
(I did it based on my local history)

koeppea:
Yes I saw this as well. Seemed to slipped through. But had no more time to fix this yesterday evening.
Will fix this today afternoon.

koeppea:
Should be fixed now.

koeppea:
Hi @LocutusOfBorg 
have you had a chance to look into the SEGFAULT of etterlog yet?

koeppea:
GDB seem not to be the right tool in this case - printf's seem to be better here...:
I'inserted fprintf's in *el_main.c* and *ec_passive.c*. The following snip of a GDB session shows that the global variables are different even though GDB thinks they are the same.
```
(gdb) b el_main.c:43
Breakpoint 1 at 0x406770: file /home/koeppea/dev/ettercap.locutusofborg/utils/etterlog/el_main.c, line 43.
(gdb) b ec_passive.c:95
Breakpoint 2 at 0x7ffff7b7cde8: file /home/koeppea/dev/ettercap.locutusofborg/src/ec_passive.c, line 95.
(gdb)  
(gdb) 
(gdb) r
The program being debugged has been started already.
Start it from the beginning? (y or n) y

Starting program: /home/koeppea/tmp/ettercap/bin/etterlog ettercap-logfile.eci
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

etterlog 0.8.2 copyright 2001-2015 Ettercap Development Team


Breakpoint 1, main (argc=2, argv=0x7fffffffe6b8)
    at /home/koeppea/dev/ettercap.locutusofborg/utils/etterlog/el_main.c:43
43	   fprintf(stderr, "el_main.c: GBL_CONF(%p)\n", GBL_CONF);
(gdb) n
el_main.c: GBL_CONF(0x61b7f0)
46	   SAFE_CALLOC(GBL_TARGET, 1, sizeof(struct target_env));
(gdb) print GBL_CONF
$1 = (struct el_conf *) 0x61b7f0
(gdb) c
Continuing.
Log file version    : 0.8.2
Timestamp           : Thu Jun 18 22:21:39 2015 [542686]
Type                : LOG_INFO

1766 tcp OS fingerprint

20530 mac vendor fingerprint

2182 known services



==================================================
 IP address   : 2.16.218.19 
 Hostname     : www.abc.de 

Breakpoint 2, print_host (h=0x628e60) at /home/koeppea/dev/ettercap.locutusofborg/src/ec_passive.c:95
95	   fprintf(stderr, "ec_passive.c: GBL_CONF(%p)\n", GBL_CONF);
(gdb) n
ec_passive.c: GBL_CONF(0x2e382e300030e77e)
96	   if (GBL_CONF->geoip_support_enable)
(gdb) print GBL_CONF
$2 = (struct el_conf *) 0x61b7f0
(gdb) 
```

In the el_* context, *GBL_CONF* has the address **0x61b7f0** in memory while in the ec_* context, *GBL_CONF* has the address **0x2e382e300030e77e** in memory. I think this comes due to linking ettercap against libettercap some time ago.

However still I don't yet know how to overcome this....

koeppea:
Maybe by calling **globals_alloc()** from libettercap. This would require to parse the configuration file also. But this should be a bad thing..... let me try this...

LocutusOfBorg:
Yes, I did many tries, and I came to a conclusion:
in etterlog/etterfilter, we use the same gbls, but with different content, we try to use libettercap functions and we succeed, but just because e.g. the functions used do not call any DEBUG or LOG function.  (because they are undefined in the context of the library).

What we are trying to do is to use functions from libettercap without properly initialization.

The solution might be:
* stop using the fake calls to printf for debug and use the libettercap version
* rename the gbls to some other thing, and use *also* the gbls from libettercap (the one you need BTW)
* create a "libettercap_init" with the proper inizialization code
* use "extern" variables to pass something to libettercap when we initialize (e.g. the pointers to something we use in the main)
* use libettercap code with no problems at all
* remove duplicated code (e.g. el_target.c) I tried to do it once if you remember, but I was stopped by the above points.

I think we need to do all of this to have a real libettercap library and usable.

what do you think?

koeppea:
That sounds good. I suggest doing this with a dedicated pull request and merging in the right sequence.
Do you agree?

LocutusOfBorg:
Yes, I agree, unfortunately this is a lot of work, I'll try to trace it down :)
If we can close somewhere this pull request and merge we can follow up later

koeppea:
Yes I agree. Let me put the Geo IP info also into the XML printout. Then it's ready to go. Will do it today evening.

koeppea:
OK done. Now I'd say it's ready for merge.

LocutusOfBorg:
so let's start another branch with the libettercap fixes :)

thanks @gvanem for your contribution!

PR #681, opened 2015-05-06
DNS TTL customizability for dns_spoof plugin

frekky:
This adds an option to all entries in etter.dns which allows the spoofed DNS record's TTL/expiry time to be set to any arbitrary value within the DNS spec (see RFC 2181 for details).

This has been tested locally using a few different record types, although it's entirely possible I omitted something.

If TTL field isn't specified in the etter.dns file, it should default to 3600 for that entry.

### DIFF
--- a/plug-ins/dns_spoof/dns_spoof.c
+++ b/plug-ins/dns_spoof/dns_spoof.c
@@ -101,16 +101,16 @@ int plugin_load(void *);
 static int dns_spoof_init(void *);
 static int dns_spoof_fini(void *);
 static int load_db(void);
-static int parse_line(const char *str, int line, int *type_p, char **ip_p, u_int16 *port_p, char **name_p, unsigned int *ttl_p);
+static int parse_line(const char *str, int line, int *type_p, char **ip_p, u_int16 *port_p, char **name_p, u_int32 *ttl_p);
 static void dns_spoof(struct packet_object *po);
 static int prepare_dns_reply(u_char *data, const char *name, int type, int *dns_len, int *n_answ, int *n_auth, int *n_addi);
-static int get_spoofed_a(const char *a, struct ip_addr **ip, unsigned int *ttl);
-static int get_spoofed_aaaa(const char *a, struct ip_addr **ip, unsigned int *ttl);
-static int get_spoofed_txt(const char *name, char **txt, unsigned int *ttl);
-static int get_spoofed_ptr(const char *arpa, char **a, unsigned int *ttl);
-static int get_spoofed_mx(const char *a, struct ip_addr **ip, unsigned int *ttl);
-static int get_spoofed_wins(const char *a, struct ip_addr **ip, unsigned int *ttl);
-static int get_spoofed_srv(const char *name, struct ip_addr **ip, u_int16 *port, unsigned int *ttl);
+static int get_spoofed_a(const char *a, struct ip_addr **ip, u_int32 *ttl);
+static int get_spoofed_aaaa(const char *a, struct ip_addr **ip, u_int32 *ttl);
+static int get_spoofed_txt(const char *name, char **txt, u_int32 *ttl);
+static int get_spoofed_ptr(const char *arpa, char **a, u_int32 *ttl);
+static int get_spoofed_mx(const char *a, struct ip_addr **ip, u_int32 *ttl);
+static int get_spoofed_wins(const char *a, struct ip_addr **ip, u_int32 *ttl);
+static int get_spoofed_srv(const char *name, struct ip_addr **ip, u_int16 *port, u_int32 *ttl);
 char *type_str(int type);
 static void dns_spoof_dump(void);
 
@@ -124,7 +124,7 @@ struct plugin_ops dns_spoof_ops = {
     /* a short description of the plugin (max 50 chars) */                    
    .info =              "Sends spoofed dns replies",  
    /* the plugin version. */ 
-   .version =           "1.32",
+   .version =           "1.3",
    /* activation function */
    .init =              &dns_spoof_init,
    /* deactivation function */                     
@@ -233,7 +233,6 @@ static int load_db(void)
       d->port = port;
       d->text = NULL;
       d->ttl = ttl;
-      USER_MSG("dns_spoof: added entry! IP %s, TTL %d\n, d->ttl %d", ip, ttl, d->ttl);
 
       /* convert the ip address */
       if (type == ns_t_txt) {
@@ -264,7 +263,7 @@ static int load_db(void)
 /*
  * Parse line on format "<name> <type> <IP-addr> <ttl>".
  */
-static int parse_line(const char *str, int line, int *type_p, char **ip_p, u_int16 *port_p, char **name_p, unsigned int *ttl_p)
+static int parse_line(const char *str, int line, int *type_p, char **ip_p, u_int16 *port_p, char **name_p, u_int32 *ttl_p)
 {
    static char name[100+1];
    static char ip[MAX_ASCII_ADDR_LEN];
@@ -277,7 +276,7 @@ static int parse_line(const char *str, int line, int *type_p, char **ip_p, u_int
    /* Set default TTL of 1 hour if not specified */
    ttl = 3600;
 
-   if (sscanf(str,"%100s %10s %40[^\r\n# ] %u", name, type, ip, &ttl) < 3) { /* TTL should be optional - this should work */
+   if (sscanf(str,"%100s %10s %40[^\r\n# ] %u", name, type, ip, &ttl) < 3) { /* TTL is optional therefore only require 3 things here */
       USER_MSG("dns_spoof: %s:%d Invalid entry '%s'\n", ETTER_DNS, line, str);
       return (0);
    }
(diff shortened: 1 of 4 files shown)


plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -116,7 +121,7 @@ struct plugin_ops dns_spoof_ops = {
     /* a short description of the plugin (max 50 chars) */                    
    .info =              "Sends spoofed dns replies",  
    /* the plugin version. */ 
-   .version =           "1.2",   
+   .version =           "1.32",
```
LocutusOfBorg: maybe you were referring to 1.3 there? :)

frekky:
Ah yes - that was just for testing purposes to make sure the build and install worked properly. Right now its quite late so I'm about to stop, but I'll try to fix any issues when I can :)

plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -251,21 +259,29 @@ static int load_db(void)
...
-      USER_MSG("dns_spoof: %s:%d Invalid entry %s\n", ETTER_DNS, line, str);
+   /* Set default TTL of 1 hour if not specified */
+   ttl = 3600;
+
+   if (sscanf(str,"%100s %10s %40[^\r\n# ] %u", name, type, ip, &ttl) < 3) { /* TTL should be optional - this should work */
+      USER_MSG("dns_spoof: %s:%d Invalid entry '%s'\n", ETTER_DNS, line, str);
       return (0);
    }
+   
+   if (ttl > 2147483647) ttl = 3600; /* keep TTL within DNS standard limits (2^31 - 1) - see RFC 2181 */
```
koeppea: for a better readability you should define MAX_DNS_TTL as 1 << 31 - 1 at the top using preprocessor and refer it in the if-statement.

plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -490,6 +506,18 @@ static void dns_spoof(struct packet_object *po)
...
+ */
+static u_int32 little_to_big(u_int32 n)
+{
+	/* Convert TTL to big-endian */
+	return ((n>>24)&0x000000ff) | // move byte 3 to byte 0
+					 ((n<<8)&0x00ff0000) | // move byte 1 to byte 2
+					 ((n>>8)&0x0000ff00) | // move byte 2 to byte 1
+					 ((n<<24)&0xff000000); // byte 0 to byte 3
+}
+
```
koeppea: What is this *little_to_big()* function about? 
Usually you just have to convert host-ordered bytes to network-ordered bytes using the standard function **htonl()**.

koeppea:
Please also update the ettercap-plugins man-page with a bit more descriptive text.

frekky:
Added a bunch of fixes for minor issues. Man page is now updated and it is definitely dns_spoof version 1.3 now.

koeppea:
Yeah better now, but still have only some minor cosmetic things to address... (see inline comments).

plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -19,6 +19,11 @@
 
 */
 
+/*
+ * May 2015
+ * Frekky - Added TTL option for spoofed DNS replies.
+*/
+
```
koeppea: Don't you think your credits would be better placed within the AUHTORS file rather than in the source code?

plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -251,21 +261,29 @@ static int load_db(void)
...
 
- DEBUG_MSG("%s:%d str '%s'", ETTER_DNS, line, str); 
+   DEBUG_MSG("%s:%d str '%s'", ETTER_DNS, line, str); 
+
+   /* Set default TTL of 1 hour if not specified */
+   ttl = 3600;
 
-   if (sscanf(str,"%100s %10s %40[^\r\n# ]", name, type, ip) != 3) {
-      USER_MSG("dns_spoof: %s:%d Invalid entry %s\n", ETTER_DNS, line, str);
+   if (sscanf(str,"%100s %10s %40[^\r\n# ] %u", name, type, ip, &ttl) < 3) { /* TTL is optional therefore only require 3 things here */
```
koeppea: Could you please put the comment above. Wherever possible the line length shouldn't exceed 80 columns. However sometimes this rule is broken in favor of the readability of the code (e.g. prototypes).

plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -503,6 +524,7 @@ static int prepare_dns_reply(u_char *data, const char *name, int type, int *dns_
    struct rr_entry *rr;
    bool is_negative;
    int len;
+   u_int32 ttl = 3600, ttl_bige = 0x10e00000; /* 1 hour by default in case something goes wrong / 3600 big-endian */
```
koeppea: Please separate the declaration from the defenition as defined in C99 standard AFAIR.

plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -545,12 +567,15 @@ static int prepare_dns_reply(u_char *data, const char *name, int type, int *dns_
          len = 12 + IP_ADDR_LEN;
          SAFE_CALLOC(answer, len, sizeof(u_char));
          p = answer;
-      
+
+         /* make TTL big-endian to work with DNS */
+         ttl_bige = htonl(ttl);
```
koeppea: I don't really see the need for the *ttl_bige* variable.
The host-byte-ordered value just need to be rewritten to the network-byte-ordered value. Not big- or little-endian
This can be done be reusing the variable: 
```c
ttl = htonl(ttl);
```

koeppea:
I've tested your code on Debian Wheezy, Jessie and FreeBSD. It works on all platforms as expected. Good job! 
However, I get some warnings, less detailed on Wheezy with gcc but more detailed on Jessie with gcc and on FreeBSD with clang:
```
plug-ins/dns_spoof/dns_spoof.c: In function ‘parse_line’:
plug-ins/dns_spoof/dns_spoof.c:42:31: warning: integer overflow in expression [-Woverflow]
 #define MAX_DNS_TTL (1 << 31) - 1     /* Maximum DNS TTL according to RFC 2181 */
                               ^
/home/koeppea/dev/ettercap/plug-ins/dns_spoof/dns_spoof.c:284:14: note: in expansion of macro ‘MAX_DNS_TTL’
    if (ttl > MAX_DNS_TTL) ttl = 3600; /* keep TTL within DNS standard limits (2^31 - 1) - see RFC 2181 */
              ^
plug-ins/dns_spoof/dns_spoof.c:42:31: warning: integer overflow in expression [-Woverflow]
 #define MAX_DNS_TTL (1 << 31) - 1     /* Maximum DNS TTL according to RFC 2181 */
                               ^
plug-ins/dns_spoof/dns_spoof.c:337:17: note: in expansion of macro ‘MAX_DNS_TTL’
       if (ttl > MAX_DNS_TTL) ttl = 3600; /* keep TTL within DNS standard limits (2^31 - 1) - see RFC 2181 */
                 ^
plug-ins/dns_spoof/dns_spoof.c: In function ‘get_spoofed_aaaa’:
plug-ins/dns_spoof/dns_spoof.c:1075:74: warning: unused parameter ‘ttl’ [-Wunused-parameter]
 static int get_spoofed_aaaa(const char *a, struct ip_addr **ip, u_int32 *ttl)
```
Seem to have something to to do with my suggestion defining the max value as  a pre-processor definition.... we're better of using a pre-defined macro for that
This works thogh:
```c
#define MAX_DNS_TTL INT_MAX
```

The last warning is about a case where ttl is unused. If it's really the case, please tell the compile that the variable isn't used by casting it to **void**:
```c
/* variable not used in this block */
(void) ttl;
```

plug-ins/dns_spoof/dns_spoof.c
frekky: TTL is used later to log when DNS replies are sppofed and print out TTL, it may be simpler to use a separate variable rather than a call to `ntohl()` (in case system is little-endian). Otherwise it may be possible to simply use `htonl(ttl)`in the memcpy call rather than `ttl_bige`

koeppea: Ah yes you are right. Then the second variable is necessary (even though I don't know the word "bige"). memcpy requires pointers so that a direct call of htonl within memcpy cannot do that.

frekky: bige = bige(ndian)  ;)

Perhaps I should change that to something more clear?

frekky:
It seems GCC doesn't understand unused variables very well. The TTL for AAAA records wasn't being set properly.

All better now :)

plug-ins/dns_spoof/dns_spoof.c
koeppea: Ah. I also don't have the gold idea. Maybe *dns_ttl*?

koeppea:
> It seems GCC doesn't understand unused variables very well.

It's just a warning, not an error. It tells you, that this could be something you forgot to implement, which seemed to be the case actually :wink: 
The second reason is, if it's really unused, that you tell the compiler that this variable doesn't need to appear in the Assembler code within this block. Otherwise it appears in the Assembler code w/o any use which is less optimal obviously.

@LocutusOfBorg @eaescob Your opinion for this pull?

plug-ins/dns_spoof/dns_spoof.c
```diff
@@ -503,9 +521,14 @@ static int prepare_dns_reply(u_char *data, const char *name, int type, int *dns_
    struct rr_entry *rr;
    bool is_negative;
    int len;
+   u_int32 ttl, ttl_bige;
    u_char *answer, *p;
    char tmp[MAX_ASCII_ADDR_LEN];
 
+   /* set TTL to 1 hour by default or in case something goes wrong */
+   ttl = 3600;
+   ttl_bige = 0x10e00000; /* big endian for network stuff */
```
LocutusOfBorg: is this equivalent to
```ttl_bige = htonl(ttl);```

if yes I would maybe prefer the htonl, less efficiency but more readability....

plug-ins/dns_spoof/dns_spoof.c
LocutusOfBorg: yes, maybe dns_ttl is better... :)

LocutusOfBorg:
I tested and everything seems good so far

frekky:
Some minor refactoring, ``` ttl_bige ``` is now ```dns_ttl``` :)

eaescob:
:+1: good work guys!

PR #683, opened 2015-06-17
Update curl

LocutusOfBorg:
and LuaJIT


koeppea:
Don't know how to test LuaJIT.
Tested curl by compiling ettercap with bundled curl and tested with SSTStrip plugin using my SSLStrip test-website.

Works as expected.

PR #699, opened 2015-08-24
wait-loop for reading data of sslstrip plugin

koeppea:
This pull request fixes a sporadic issue with **sslstrip**.

The following prerequisites have to be met:
 1. traffic redirect is active
 2. sslstrip plugin is active
 3. delay between ettercap and the victim or a utilized victim's system

Under these circumstances, the delay between the last ACK of the 3-way handshake and the first PSH,ACK containing the HTTP request may be longer than ettercap needs to process the *read()* call of the currently established connection. 

As a consequence of the non-blocking treatment of the socket file descriptor, *read()* immediately exits with a negative return code.

Due to improper evaluation of the error-code, the connection is wiped (FIN) before the first ACK with data can arrive at ettercap. The result of this race-condition is that ettercap's host rejects this request packet with a RST, because it already finished (FIN) the connection. On the victim's machine it looks like the connection is sometimes rejected and a "*reload*" sometimes helps.

I figured this out while I tested something in parallel while testing something as the victims machine.
Not only system utilization can trigger this bug, also link latency can fail this race-condition which can be found often in WiFi networks.

Hope I explained it somewhat understandable.

### DIFF
--- a/plug-ins/sslstrip/sslstrip.c
+++ b/plug-ins/sslstrip/sslstrip.c
@@ -793,16 +793,37 @@ static int http_sync_conn(struct http_connection *connection)
 
 static int http_read(struct http_connection *connection, struct packet_object *po)
 {
-   int len = 0;   
+   int len = 0, ret = -E_INVALID;
+   int loops = HTTP_RETRY;
 
-   len = read(connection->fd, po->DATA.data, HTTP_MAX);
+   do {
+      len = read(connection->fd, po->DATA.data, HTTP_MAX);
+
+
+      if(len <= 0) {
+         /* in non-blocking mode we have to evaluate the socket error */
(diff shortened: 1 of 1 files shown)


LocutusOfBorg:
+1 good catch!!! I got this bug I guess many times, just a question: what about src/ec_sslwrap.c:865 code? does it have the same issue?

koeppea:
In this read routine this distinguishment is already implemented.
Furthermore is this not that a big deal as ettercap doesn't need to wait for the first GET or POST as it's right away trying to establish the SSL connection to the target server, as it's deriving information of the server's certificate for the fake certificate.

LocutusOfBorg:
merged thanks :)

Issue #471, opened 2014-01-17
Problem with ettercap


ghost:
Today, when I run the ettercap to sniff the facebook on my local network. I have a big problem that the facebook said that "Cannot connect to the real www.facebook.com". So that I think this trick doesn't work anymore. Do anyone are the same with me?

I have tested on: IE, Firefox, Chrome and all of them are get me to the same situation.

PR #701, opened 2015-09-04
SSLwrap fixes

koeppea:
This pull request tried to address 3 issues in regard to SSL interception.
 1. When a extablished SSL browser session is established and then you turn on ettercap with SSL interception, then all further SSL connection attemps are rejected because it's just unlikely that the issuer change but not the serial of the certificate. The first change addresses the issue and sets a ettercap custom serial number for the certificate to overcome this detection
 2. The fake self-singned certificed already expired ages ago. The new certificate just uses SHA1 instead of MD5 as a signature algorithm and extends the validitiy so that it could potentially added to the trusted certificates store.
 3. Intercepting SSL connection to sites s.a. www.facebook.com which was already reported with #471. But my research revealed that this is due to the RFC6797 this isn't that easily possible. Especially because it revealed that Firefox & Co. has a built-in HSTS hosts list where www.facebook.com is also included. The only way to circumvent this is to make the victim make accept a fake certificate for www.facebook.com and redirect to a rouge webserver hosting or proxying the actual site.

Due to the inability to influence the built-in HSTS cache of a victim's browser,  only point 1 and 2 are addressed within pull request.

However, I'm still not satisfied with point 3. I cannot stop thinking of howto circumvent this approach and finally using ettercap.


LocutusOfBorg:
merging thanks!

koeppea:
@LocutusOfBorg What @gvanem mentiones is something disregard to this pull request.
@gvanem The recommendation on any Linux distribution is to use the packages from the distribution. So Debian's package repository is widely used even by other distributions. Of course the current stable source packages are not equal to the latest one from upstream. With the ettercap package, it behaves exactly the same.

I'll try to reproduce on a Linux system using the latest source from upstream and make up a fix that covers both, the currently rolled out and the latest versions of OpenSSL.

koeppea:
> This file fails to compile here with a recent OpenSSL:

I can confirm this.
I'm working on it... opening a new pull request when fixed.

PR #703, opened 2015-09-08
502 fix

LocutusOfBorg:


koeppea:
Ain't we're better off just returning from the function rather than using the BUG_IF macro which aborts the complete execution of ettercap. In the middle of a MITM session it disconnects the targets until they refresh their neighbor cache.
What do you think?

LocutusOfBorg:
thanks @koeppea actually I wondered about the case, and yes, you are right!

I guess I updated it correctly :)

koeppea:
:+1:

LocutusOfBorg:
Merged thanks!

Issue #748, opened 2016-08-13
etterlog: memory leak when processing crafted file


asarubbo:
There is a memory leak when processing crafted file.

~~~
# etterlog $FILE
Log file version    : 0.8.2
Timestamp           : Thu Jul 16 15:28:54 2015 [688192]
Type                : LOG_INFO

1766 tcp OS fingerprint

20530 mac vendor fingerprint

2182 known services


=================================================================
==13629==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 49321 byte(s) in 1 object(s) allocated from:
    #0 0x4c1ae0 in calloc /var/tmp/portage/sys-devel/llvm-3.8.1/work/llvm-3.8.1.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:66
    #1 0x502af9 in get_info /var/tmp/portage/net-analyzer/ettercap-9999/work/ettercap-9999/utils/etterlog/el_log.c:207:7

Direct leak of 8232 byte(s) in 1 object(s) allocated from:
    #0 0x4c1ae0 in calloc /var/tmp/portage/sys-devel/llvm-3.8.1/work/llvm-3.8.1.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:66
    #1 0x508a16 in globals_alloc /var/tmp/portage/net-analyzer/ettercap-9999/work/ettercap-9999/utils/etterlog/el_main.c:132:4

Direct leak of 232 byte(s) in 1 object(s) allocated from:
    #0 0x4c1958 in malloc /var/tmp/portage/sys-devel/llvm-3.8.1/work/llvm-3.8.1.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:52
    #1 0x7f716a4879c9 in gz_open /var/tmp/portage/sys-libs/zlib-1.2.8-r1/work/zlib-1.2.8-abi_x86_64.amd64/gzlib.c:111

Direct leak of 5 byte(s) in 1 object(s) allocated from:
    #0 0x4c1ae0 in calloc /var/tmp/portage/sys-devel/llvm-3.8.1/work/llvm-3.8.1.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:66
    #1 0x502883 in get_info /var/tmp/portage/net-analyzer/ettercap-9999/work/ettercap-9999/utils/etterlog/el_log.c:199:7

Indirect leak of 16384 byte(s) in 1 object(s) allocated from:
[...]

PR #757, opened 2016-09-24
Free up memory to address leak reported in #748

eaescob:


eaescob:
Can you try again with these changes?

PR #761, opened 2016-10-12
Improving .gitignore

RyDroid:


.gitignore
```diff
@@ -1,3 +1,60 @@
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided this notice is
+# preserved.  This file is offered as-is, without any warranty.
+# Names of contributors must not be used to endorse or promote products
```
LocutusOfBorg: This seems to be not english :) and I would prefer avoiding this at all, we already have a global license

RyDroid: I don't get your point about english. It is a mix of the GNU All-Permissive License and the 3-clause BSD license. https://www.gnu.org/prep/maintain/html_node/License-Notices-for-Other-Files.html https://directory.fsf.org/wiki/License:BSD_3Clause

LocutusOfBorg: oh well, it was fine english-wise, but I don't like additional custom licenses, the main gpl-2-ish license of ettercap should be fine for a gitignore file too :)
(and as a Debian maintainer I really don't like having to patch my copyright file just for it)

Issue #769, opened 2016-11-10
Possible memory leak or unwanted behavior in USER_MSG from plugin


mkslavin:
Here is a proof of concept plugin that will inevitably fill the memory on a computer if it runs long enough  (collects enough packets) and will trigger the OOM killer. https://gist.github.com/mkslavin/a11b7baa468eeb3f8b1eaae7312604f3

If you run this plugin with ettercap running in text mode with the valgrind tool Massif, you can see that ui_msg is grabbing a lot of memory and never releasing it. Here is sample output from running on my machine https://gist.github.com/mkslavin/f4ffab499767177b643aa624335d8dd4 . The more frequently USER_MSG is called, the faster the memory usage grows. Using printf instead resolves this issue but then breaks compatibility with non-text mode ettercap UIs.

koeppea:
Looking at the code in the light of your plugin experiment, I don't see this being a memory leak.
What you describe and experience is by design of how the text interface works in ettercap and how you exploit this behaviour by your plugin code.

The text user interface acts on a 10ms interval basis if no user (keystroke) input is given. At each interval the messages buffer is freed by 10 messages which are then displayed.
Your plugin, given a moderate packet rate going over the wire, easily exceeds 10 USER_MSG calls per 10ms (or 1000 USER_MSG calls per second). As USER_MSG is not designed being a lossy printout of arbitrary data, ettercap queues all messages. The user interface is then responsible to fetch the messages off the queue. Hence queue is filled faster than emptied.

Can you try if this issue is still reproducable using my test pull request #771 ?

However, I still don't see this being a real issue since user messages are designed for the user to print useful information being able to be read by human.

mkslavin:
Sorry for the long gap in response-  commit edf3e66 fixes this issue and USER_MSG works as expected now.

PR #771, opened 2016-11-21
flush max messages per interval in text interface

koeppea:
This Pull Request is just to test if this addresses the issue #769. 
Text interface is the only one that pulls only a small amount of messages off the queue per cycle.
However, the referred issue is somewhat theoretic from my point of view as it contradicts the purpose of the messages pane.

So ATM I don't like to see this PR merged. More being a educated test.

### DIFF
--- a/plug-ins/CMakeLists.txt
+++ b/plug-ins/CMakeLists.txt
@@ -50,7 +50,8 @@ search_promisc
 smb_clear
 smb_down
 smurf_attack
-stp_mangler)
+stp_mangler
+umsg_test)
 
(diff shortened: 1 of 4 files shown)


plug-ins/CMakeLists.txt
```diff
@@ -50,7 +50,8 @@ search_promisc
 smb_clear
 smb_down
 smurf_attack
-stp_mangler)
+stp_mangler
+umsg_test)
```
LocutusOfBorg: @koeppea missing file? :)

LocutusOfBorg:
merged thanks :)

Issue #774, opened 2016-12-14
Does Plugin dns_spoof work in bridge mode?


hmn91:
Hi guys,
I am trying to run dns_spoof plugin in ettercap bridge mode. Here is my command 
`ettercap -TqS -i eth0 -B eth1 -P dns_spoof` 
and my etter.dns look like this 
`example.com   A   1.2.3.4`
eth0 connect to the internet and eth1 is connect to the victim, 1.2.3.4 is my vps ip address
Now I browse example.com from victim machine, it cannot connect to my vps. I have capture data using wireshark on victim machine and see no dns reply. nslookup it and get timeout.
So does dns_spoof work in bridge mode? Do I miss any config?

hmn91:
As I try in bridge mode, the victim machine and attacker machine cannot connect to each other, for example, in attacker machine run a local website, the victim cannot connect to it, and vice versa.
In dns_spoof, I think it drop the query and create a resonse itself. But because it cannot connect to victim machine so the response cannot be sent. Am I right?

koeppea:
thanks. I've been able to reproduce the issue.
Will further look into it.

koeppea:
@hmn91 could you please test if the PR #775 helps?
Thanks.

hmn91:
@koeppea The dns_spoof works well now. Thank you so much :).
Can you also have a look the problem I described above about in bridge mode, attacker machine and victim machine cannot connect to each other. For more details, I have something like this
Computer A (eth0: 192.168.1.3) <-------> (eth1) Computer B (eth0: 192.168.1.2) <-------> Gateway 
From computer A, I can access Internet normally, every packets is forwared well. But in case I want access, for example a web page, on computer B from computer A, it failed. As I capture the packets, they go like this:
1. A(192.168.1.3) send Get Request to B(192.168.1.2)
2. B receives the request on interface eth1
3. B broadcast ARP request: Who has 192.168.1.3? **on interface eth0**
now of course there is no reply to this ARP so B do not know how to send data to A
As I check B's arp table, it have 192.168.1.3 on eth1 but not on eth0.
That's all information I have till now. Please have a look :). Thanks.

koeppea:
Good.
Please open a new issue to keep these things separated.
Also point out on which machine ettercap is running (A or B) and on which machine the webserver is running to be able to properly reproduce.

hmn91:
@koeppea i have create new issue, please read here https://github.com/Ettercap/ettercap/issues/776

PR #775, opened 2016-12-17
Fix 774: reliable reflect spoofed DNS reply on same interface

koeppea:
This pull request addresses issue #774 as the send function does not properly take the dual interface situation into account in bridged sniffing mode.


koeppea:
Hi @LocutusOfBorg too fast.
I intended to prepare a alternative approach which in my point of view may be more elegant:
in _send_m?dns_reply()_:
 1. use libnet to build the protocol stack for the DNS reply
 2. create ettercap packet_object 
 3. use the sniffing method's forwarding function

What do you think how to approach this since this PR is now merged already?

koeppea:
@LocutusOfBorg do you know why some of the Travis builts failed? Mainly MacOSX.

LocutusOfBorg:
Do another pr :) yes there has been a Macosx regression that made builds fail.
I honestly don't care too much since I have no real hw here

koeppea:
Nevermind. the outlined approach doesn't make it more elegant, since building of the link-layer header is dependent on the outgoing interface. So this approach would just lead to additional processing.
So merging this one was correct.

LocutusOfBorg:
thanks :) it looked really nice and clean to me!

koeppea:
> since I have no real hw here

I have meanwhile a MacOS device available. I'll try to reproduce the Travis failure.

PR #779, opened 2017-01-21
Fix Travis osx build

koeppea:
This PR aims to fix Travis OSX related build errors. Crossing fingers....

Since this PR is somewhat platform dependent, there should be no impact on any other platform. However this needs to be proved.


koeppea:
@LocutusOfBorg I'm thinking about adding some descriptive information how to build ettercap on MacOSX.

Especially because a Cmake variable need to be set to determine the correct location of the OpenSSL package and libraries and I've not managed to get this automated within Cmake files. Maybe you can help there:

```
$ cmake -DOPENSSL_ROOT_DIR=`brew --prefix openssl` ..
```

What do you think would be a  place to get that documented?

LocutusOfBorg:
I still don't understand why commit fe13c7bc9782c794a3d7eabcae634bbe00a13734 should make any difference... I explicitly called the result of the two "find" functions with the same name, to avoid having duplicate flags :)

also, I don't understand 2fc6badf4abb1e45c4a4cabaa8a6296a4bdd7615 but I trust you :)

ad6c5ae7415195e4c263b47ee867000977c3ba92 this one seems good, even if I don't understand where is the fault (ettercap, brew, slowlyness in travis?)

koeppea:
> I still don't understand why commit fe13c7b should make any difference... I explicitly called the result of the two "find" functions with the same name, to avoid having duplicate flags :)

the main issue why the _tests_ target failed to link check using `-lcheck` was due to the missing **LDFLAGS** not used which the PkgConfig find functions determined though. I've distinct both variable prefixes to be more explicit. Honestly, I didn't thought that it was by intention.


> also, I don't understand 2fc6bad but I trust you :)

I've seen a warning when I performed the pre-requisite installation manually on my MacBook. I think since it's exclusive to MACOSX it won't harm other platforms.



> ad6c5ae this one seems good, even if I don't understand where is the fault (ettercap, brew, slowlyness in travis?)

This is actually because brew need to install _openssl_ first, after that it can determine the installation prefix (`--brew --prefix`), which is needed to give **cmake** a hint otherwise it just doesn't finds the library. The ugly thing about that is, that didn't manage to automate this (_execute `brew --prefix openssl` to set cmake variable_) using cmake itself (chicken/egg). Do you have idea? If not, where to document this caveat in the build instructions?

eaescob:
I need to buy myself a new macbook pro. What's missing for this PR to be merged?

LocutusOfBorg:
fine for me! +1

koeppea:
Maybe we have to rebase some of the failed PRs over these changes to make Travis happy.

PR #804, opened 2017-06-08
Asan support is not intended for production, disable it

LocutusOfBorg:
http://www.openwall.com/lists/oss-security/2016/02/17/9


koeppea:
Stange. I thought I did append the flags to the _DEBUG Cmake variable.

LocutusOfBorg:
yes, but the problem is that many distro use RelWithDebInfo as target (release with debug stuff), because this way we can objdump strip the debug symbols and release the -dbg version of the package

PR #797, opened 2017-04-28
Deprecate inet_aton

koeppea:
This pull request is more a janitor work to deprecate the use of `inet_aton()` but also fixing a minor undiscovered bug and enhancing the gre_relay plugin.

@LocutusOfBorg if this PR is too heavily interfering with your library fix branch, it can wait. No need for extra work. I'll rebase it once your library fix has made it into **master**.

PR #814, opened 2017-07-23
Dedicate own source file for inet_pton implementation

koeppea:
This PR is in response to @gvanem 's request (#797) to move the missing *inet_pton()* implemenation to an own source file and include it conditionally if a system provided implementation exists or not, regardless which operating system is used for the build.

@gvanem Could you please test if it works as desired?


gvanem:
At first glance, it look okay to me (not built with it).

koeppea:
I'd be interested in build and run.

sgeto:
the cmake test will return "not found" on windows, yet the lack of inet_pton is taken care of by winsock, which is needed by ettercap on windows anyways. it seems to me that linking (lib)ettercap with winsock 2 is all it takes.
I can't think of a scenario where this isn't the case and a local inet_pton.c is in order.
How would I test for this?

koeppea:
I don't know, since I don't have a Windows system.
I wasn't aware of another winsock library that implements that. Is something special necessary to link against winsock2?
When I understand you correctly, I should conditionally omit the cmake function-check on Windows. Correct?

However, I don't know how up2date the **README.PLATFORMS** is in regard to the compilation hints on Windows. An update to that is well appreciated.

sgeto:
When I said winsock, I meant winsock2. You can't link with both at same time. Ettercap's sources are already including its header (winsock2.h), but cmake isn't set up to link with the library (ws2_32).

I think that, with respect to windows vista and above, none of the changes you are sugesting in this PR are needed. Except for the lines you removed in src/os/ec_mingw.c. Those are useless. The rest is taken care of by winsock2.
Again, I'm referring to vista and above. It may different on older versions of windows.
@gvanem Am I missing something?

Without going into too many details, the Windows port has fallen behind quite a lot. No one will be able to easily test your PR and build against it on your master.

Updating README.PLATFORMS alone won't solve this. The source is mostly "alright", but the build system needs a complete overhaul and extensive testing. For MinGW, that is. MSVC is another story, although I very much like to see that working out.

Whenever i find the time, I am working on making that happen on a separate branch of my fork (https://github.com/sgeto/ettercap/tree/win32). 
cmake and I are immortal enemies. I think it will take me a few more weeks before I can hand in a PR. 
Sooo... stay tuned.

gvanem:
@sgeto I don't use or like Cmake, so I don't know about it's state in Ettercap. But as you stated, `inet_ntop()` is for sure not present in `ws2_32.dll` on Win-XP's. But everybody should ignore that OS.

sgeto:
Ah, so it's true. I agree xp is EOL. People should stay away from it.

What this PR is concerned, I don't think there's any reason have this. Just for xp?

koeppea:
Alright. I've no issues in reducing that PR to the removal of unnecessary code.
The local maintenance of inet_pton was just for parity reasons.

LocutusOfBorg:
I agree, we shouldn't support windows xp.

@sgeto if you have cmake issues, I'm happy to help out, as long as you post me build errors and don't ask to test stuff on windows
(I have a windows 10.1 virtual machine, I might try to help, if you provide documentation on how to do it :) )

sgeto:
Thank you @LocutusOfBorg

I will definitely need help to get this to work. Not so much with cmake, but with the source.
There are a many things needed to take care of. Ettercap's source is very extensive and my C programming knowledge is quite basic. I often don't know where look when I have problems.
Currently, it's telling me that the dns disector (and dns disector only >:-( ) is not supported:

screenshot 35

Excuse me, but how (and where) is that determined? I mean, how would I debug this? The ettercap*.log is not very helpful.

Another thing is that the sources don't have a "common" header file. A *.h file that is directly or indirectly included by all *.c files. There's ec.h and config.h, but many *.c files don't care about it. This is forcing me to write many conditionals and to #include or #define things over and over again, effectively spamming the ettercap source. 

The none GNU MSVC compiler for instace doesn't know anything about visibility settings, which are used all over ettercap. if there was a common header I could add something like this to it:

```
#ifdef _MSC_VER
#define __attribute__(A) /* do nothing */
#endif
```

which would take care of that. Maybe there's another way. For now, I have to add this to the top many many source files :-(

It would be very helpful if this could be done. Maybe you can add this to your library fix?

Don't worry, I will document how to build and test it on windows. Probably next weekend along with my pull request.

sgeto:
To answer my own question: the dns dissector was only build when the cache variable HAVE_DN_EXPAND is true. As a rule of thumb: if it's not in ettercap*debug.log, it has to do with cmake.
this is fun! (not really)

gvanem:
Hi, trying to build using MSVC-2015 would be brave (and futile) since `cl v1900` at-least is not really a C99 compiler; it doesn't understand variable-sized locals like:
```c
int ec_foo (size_t size)
{
  a_type x [size];
```
Tons of these cases which could have been written as:
```c
int ec_foo (size_t size)
{
  a_type *x = alloca (size*sizeof(a_type*));
```
There are other C99 isues too. Better use `clang-cl` for now.

But the `HAVE_DN_EXPAND` should really be on. Since `dn_expand()` is a Unixism from `<resolv.h>`. It get defined as: `#define dn_expand(m,e,c,ex,l)  ec_win_dn_expand (m, e, c, ex, l)`
(look in `"ìnclude/ec_os_mingw.h"` and  `"src/os/ec_os_mingw.c"`)

The lack of a common file can be fixed by forcefully including (a tweaked hand-edited) `config.h` by:
```
-FI./config.h
```
That's what I typically do for MSVC/clang-cl when autoconfig/Cmake ignores those compilers.

koeppea:
I don't know off-hand how many such variable size stack declarations ettercap's code has (I can't recall any)... 
But to get back on-topic:
Do I understand it correctly, that I should change this PR in just removing the inet_pton code from *ec_mingw.c*?

gvanem:
> I don't know off-hand how many such variable size stack declarations ettercap's code

In `src/dissectors/ec_login.c` for example:
```
      char str[strlen(s->data) + PACKET->DATA.disp_len + 2];
```
And lots more.

> Do I understand it correctly, 

I think so, yes. But someone else (Ali) could verify.

koeppea:
> In src/dissectors/ec_login.c for example:

This should be fixed of course. Would you volunteer?



> Do I understand it correctly,
> 
> I think so, yes.

Then I'll do up and rebase the PR accordingly.

koeppea:
So now, just the `inet_pton()` code in *ec_mingw.c* is being removed.

sgeto:
Sorry for the late reply guys.

@koeppea Yes, you understood correctly. 
I did another test now and it seems to work fine. Had no troubles building and executing on mingw32 and mingw64 (gcc 6.3) yet.

Thanks a lot @gvanem

I completely agree with dn_expand. The function test turns outs as false, but I'll set HAVE_DN_EXPAND to true on windows anyways.

I understand that by trying to add MSVC compatibility, I set a very high goal. As you said, maybe too high. 
cl is a great compiler. It's fast and simple. I also prefer the IDE over my text editor.
With msvc, the plan right now is to get cmake to spit out usable solution/project/nmakefile. The rest I would then just try to adress one build error at a time.

MinGW is still the priority.

Gisle, I have one more questions if you don't mind.

User input in the text interface isn't working for me  doesn't work for me.
With user input, I mean being able to hit "P" for plugins and "Q". Any idea, why? Is there something unusual to consider?

gvanem:
> User input in the text interface isn't working for me.

AFAICR, all keypresses are detected by `ec_poll_in (0, timeout)`. Which trickles down to `_win_kbhit()`.
If it doesn't, try adding some trace-code in `ec_win_poll()`. The `-T` option work fine here.

sgeto:
thanks for the hint. I thought this was related to stuff in `ec_daemon.c` which I had troubles with. No `setsid` on windows, I locally defined `SIGTTOU`, `SIGTTIN`, `SIGTSTP` and `SIGHUP` and `HAVE_DAEMON` is never defined, not even on linux. I decided to completely ignoring/skip that file now. 

I also had a look at a copy of your `ec_mingw.c` that you gave me a while ago. Maybe you remember.
Things look very different there compared to master:
```
static int __inline win_kbhit (void)
{
  int rc = 0;

#if defined(HAVE_NCURSES) && !defined(BUILDING_UTILS)
  if ((current_screen.flags & WDG_SCR_INITIALIZED))
     return PDC_check_bios_key();
#endif
  if (_kbhit())
     rc = _getch();
  if (rc)
     ungetc (rc, stdin);
  return (rc);
}

/*
 * A poll() using select()
 */
#define HOW_MANY(x, y)  ((x + (y - 1)) / y)

int ec_win_poll (struct pollfd *p, int num, int timeout)
{
  struct timeval tv;
  size_t size = FD_SETSIZE * HOW_MANY(num, FD_SETSIZE);
  fd_set *read  = alloca (size);
  fd_set *write = alloca (size);
  fd_set *excpt = alloca (size);
  int     i, n = -1, n_socks = 0, ret = 0;

  FD_ZERO (read);
  FD_ZERO (write);
  FD_ZERO (excpt);

  for (i = 0; i < num; i++) {
    if (p[i].fd < 0)
       continue;

    if ((p[i].events & POLLIN) && i != STDIN_FILENO) {
       FD_SET (p[i].fd, read);
       n_socks++;
    }

    if ((p[i].events & POLLOUT) && i != STDOUT_FILENO) {
       FD_SET (p[i].fd, write);
       n_socks++;
    }

    if (p[i].events & POLLERR)
       FD_SET (p[i].fd, excpt);

    if (p[i].fd > n)
       n = p[i].fd;
  }

  if (n == -1)
     return (0);

  if (n_socks > 0) {

    /* select() in Winsock ignores 'n+1'.
     */
    if (timeout < 0)
       ret = select (n+1, read, write, excpt, NULL);
    else {
      tv.tv_sec  = MILLI2SEC(timeout);
      tv.tv_usec = MILLI2MICRO((timeout % 1000));
      ret = select (n+1, read, write, excpt, &tv);
    }

    for (i = 0; ret >= 0 && i < num; i++) {
      p[i].revents = 0;
      if (FD_ISSET (p[i].fd, read))
         p[i].revents |= POLLIN;
      if (FD_ISSET (p[i].fd, write))
         p[i].revents |= POLLOUT;
      if (FD_ISSET (p[i].fd, excpt))
         p[i].revents |= POLLERR;
    }
  }
  else {
#if 0
    DEBUG_MSG ("%s(): n: %d, n_sock = 0, timeout: %d",
               __func__, n, timeout);
#endif
    Sleep (10);
  }

  if ((p[STDIN_FILENO].events & POLLIN) && num >= STDIN_FILENO && win_kbhit()) {
    p [STDIN_FILENO].revents = POLLIN;
    ret++;
  }
  if ((p[STDOUT_FILENO].events & POLLOUT) && num >= STDOUT_FILENO &&
      isatty(STDOUT_FILENO) >= 0) {
    p [STDOUT_FILENO].revents = POLLOUT;
    ret++;
  }
  return (ret);
}
```

And sure enough this works fine. Not only that, I am noticing much less CPU consumption using this.
I think the neverending `LOOP()` in `ec_text.c` was going crazy. **neverending loop** finally shows up in the debug log too, which is nice.
I'll use your code from here on out. Hope you don't mind.

Thanks again

gvanem:
The CPU reduction is due to the added `Sleep (10)`.
The change in `win_kbhit()` to push-back the character allowed `ec_poll_in(fileno(stdin), 10)` in `ec_text.c` etc. to work okay.

sgeto:
I see. Thanks for clarifing

LocutusOfBorg:
merged thanks! lets follow up to a new issue for the windows fixes :)

PR #825, opened 2017-08-27
Make it work with lua 2.1

LocutusOfBorg:
Lets see if this change is retro-compatible, Debian has now lua 2.1 in unstable


koeppea:
Merged. Thanks.

PR #837, opened 2017-10-11
GeoLite country database download helper script

sgeto:
A wrote this a while ago. I hope it is of use.
It's a simple shell script that helps in downloading and installing the *free* GeoLite country databases, which are needed for the GeoIP feature to actually work.
You can optionally pass an alternative install path to this script.
If you rather not, it will install all files to /usr/local/share/GeoIP/.

### DIFF
--- a/misc/geolite-update.sh
+++ b/misc/geolite-update.sh
@@ -1,0 +1,74 @@
+#!/bin/sh
+# set -xe
+
+# geolite-update.sh -- update GeoIP lite databases.
+# (C) 2017 Ettercap Development Team.
+
+# Ettercap can use MaxMind's GeoIP databases to look up the country for an IP address.
+# This simple shell script helps in downloading and installing the *free* GeoLite
+# country databases, which are needed for this feature to actually work.
+# You can optionally pass an alternative install path to this script.
+# If you rather not, it will install all files to /usr/local/share/GeoIP/.
+# Note: In some distributions/operating systems these databases are available
+# through their package manager.
+# You most likely don't need to use this script if this is the case.
+
+USAGE="USAGE: $(basename $0) [geolite install path]"
+
+# check argument count
+if [ $# -gt 1 ]
+then
+  echo $USAGE
+  exit
+fi
+
+if [ -z $1 ]
+then
+    geolite_path="/usr/share/GeoIP"
+    download_path="/usr/share/GeoIP/download"
+else
+    geolite_path=$1
+    download_path="$1/download"
+
+fi
+
+# prg="curl --remote-name"
+prg="wget --continue --directory-prefix=$download_path"
+
+if [ ! -e $geolite_path ]; then
+        echo "Unable to find GeoIP directory: $geolite_path"
+        exit 1
+fi
+
+echo "Updating/Installing GeoLite databases..."
+echo "Note: Not installing GeoLiteCity database (not used by Ettercap)"
+
+[ -d $download_path ] || mkdir $download_path
+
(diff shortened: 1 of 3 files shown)


misc/geolite-update.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/sh
+# set -xe
+
+# geolite-update.sh -- update GeoIP lite databases.
+# (C) 2017 Ettercap Development Team.
+
+# Ettercap can use MaxMind's GeoIP databases to look up the country for an IP address.
+# This simple shell script helps in downloading and installing the *free* GeoLite
```
koeppea: Wouldn't it make sense putting here some advice, that distribution provided packages are recommended and are detected automatically? This method is IMHO targeted for systems that don't provide the possiblity to easily install GeoIP database.

misc/geolite-update.sh
```diff
@@ -0,0 +1,71 @@
...
+# country databases, which are needed for this feature to actually work.
+# You can optionally pass an alternative install path to this script.
+# If you rather not, it will install all files to /usr/local/share/GeoIP/.
```
koeppea: Can you adjust the default path of the *geoip_data_file* option in the **etter.conf** file.

koeppea:
I like this helper script, 
Most Linux distributions provide packages that installs the GeoIP databases and the library finds it since it's also coming from the distribution.

But for distributions that do not provide a distribution based installation routine for geoip this script is a good helper.

misc/geolite-update.sh
sgeto: I guess you're right. I didn't notice that some distribution provide these by default.
I thought they don't since they are none-free.

misc/geolite-update.sh
sgeto: Sure.

koeppea:
Thanks looks good for me now.
Could you squeeze the commits and rebase onto lastest master?

LocutusOfBorg:
I like it, even if Debian and Ubuntu won't use it :)

koeppea:
Yeah as said. It's more aiming to some custom OS'es or even MacOS.

koeppea:
Merged. Thanks.

sgeto:
You're welcome.
I have an unrelated question:
Out of your head: Do you know if it's possible (a good idea), and what it take to update ettercap to support the newer Geoip2 library and database?
I think that this geoip feature was implemented by @gvanem so I guess he should know best.

koeppea:
Yes @gvanem made the initial proposal.
Yes it is possible. Question is what the benefit is to go for geoip2 lib? What about the availability on the various distros?
Is the current one not maintained. Does it provide anything we're lacking at the moment?

gvanem:
I heard my name here but what was the question?

>  What about the availability on the various distros?

If you mean others that GeoIP, there is the awesome [IP2Location](https://github.com/chrislim2888/IP2Location-C-Library). This includes more details for free (just a sign-up); city and region etc.

sgeto:
> Question is what the benefit is to go for geoip2 lib?
> Does it provide anything we're lacking at the moment?

I am not sure how to answer that. It's true that ettercap's geoip support is working and does what it should. I guess the fact that maxmind refers to geoip as _legacy_ and because other projects made the switch made me suggest that ettercap should do the same.
I also think that most distros provide packages to both GeoIP and GeoIP2. The switch should not pose a problem there.

@gvanem 
The question is whether ettercap should upgrade to geoip2. And if so, how much would need to be changed/adjusted codewise?

IP2Location is pretty awesome actually. If you are thinking to implement it (in favor of GeoIP), I'm happy to help.

PR #839, opened 2017-10-11
Remove the dependency on ghostscript when building the PDF docs

sgeto:
This PR _attempts_ to remove the dependency on ghostscript when building the PDF docs.
Groff can do the conversion by itself.
I think the ghostscript package is around 18 megabytes on ubuntu. On windows ghostscript is even bigger.
All PDF docs combined are less 250 kilobytes.
It's not very efficient to get such a big package to output such tiny files.

By the way: travis may error not because this doesn't work, but because there is something wrong with the groff package on precise pangolin. It's limited (or there is a bug or something). So try it on your local machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ addons:
       - check
       - cmake
       - flex
-      - ghostscript
+      - groff
       - libbsd-dev
       - libcurl4-openssl-dev
       - libgtk2.0-dev
@@ -138,7 +138,7 @@ matrix:
       env: BUILD_ARGS="-DENABLE_GTK=Off -DENABLE_CURSES=Off"
 
 before_install:
-    - if [ ${TRAVIS_OS_NAME} = "osx" ]; then brew update; brew install check curl; brew install ghostscript gtk+ gtk+3; brew install libidn libnet; brew install luajit openssl pcre rtmpdump geoip; fi
+    - if [ ${TRAVIS_OS_NAME} = "osx" ]; then brew update; brew install check curl; brew install groff gtk+ gtk+3; brew install libidn libnet; brew install luajit openssl pcre rtmpdump geoip; fi
 before_script:
     - if [ ${TRAVIS_OS_NAME} = "osx" ]; then export BUILD_ARGS="$BUILD_ARGS -DOPENSSL_ROOT_DIR=`brew --prefix openssl`";fi
 
--- a/man/CMakeLists.txt
+++ b/man/CMakeLists.txt
@@ -1,8 +1,17 @@
 # Add man page custom target
 if(ENABLE_PDF_DOCS)
-  find_program(GROFF_EXECUTABLE NAMES groff)
-  find_program(PS2PDF_EXECUTABLE NAMES ps2pdf)
-  set(PDF_FILES)
+        find_program(GROFF_EXECUTABLE NAMES groff)
+        set(PDF_FILES)
+
+# Check if groff supports PDF generation
+        execute_process(COMMAND ${GROFF_EXECUTABLE} -Tpdf -V
+        RESULT_VARIABLE GROFF_EXIT
+        ERROR_VARIABLE GROFF_ERROR
+        OUTPUT_QUIET)
+
+        if(NOT ${GROFF_EXIT} STREQUAL 0)
+          message(FATAL_ERROR "groff (at ${GROFF_EXECUTABLE}) does not seem to support PDF generation (${GROFF_EXIT}).\ngroff failed with output:\n${GROFF_ERROR}\nConsider upgrading the groff installation.\nUsers of Debian-based distributions must also install the full groff package.")
+        endif()
 endif()
 
 set (MAN_NAMES ettercap.8 etterfilter.8 etterlog.8 etter.conf.5)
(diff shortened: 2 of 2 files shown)


sgeto:
> By the way: travis may error not because this doesn't work, but because there is something wrong with the groff package on precise pangolin. It's limited (or there is a bug or something). So try it on your local machine.

Well, nevermind.

koeppea:
I like this one which is aiming to reduce the dependencies.
Unfortunately it doesn't work on my machine (Debian 9 / groff version: 1.22.3)
Make fails at:
```
[ 98%] Creating PDF for /home/koeppea/dev/ettercap/build/man/ettercap.8
/usr/bin/groff: can't find `DESC' file
/usr/bin/groff:fatal error: invalid device `pdf'
```
An idea why this fails for me but not for Travis?

sgeto:
I had the same issue on ubuntu 12.04. There is something wrong with the groff package that ships with them.
Update your package index and then reinstall it (use `apt install --reinstall` if necessary). See travis.

koeppea:
Ah. I had the package **groff-base** installed. Installing **groff** which implicitely installs **psutils** made it work.
We should somehow catch this with CMake. Maybe looking for *psutils* too.

sgeto:
Good idea

sgeto:
1. psutils isn't a single executable
which of these four does the trick?

2. I just installed psutils without updating groff-base and it refused to build.
I doubt that this package is even needed, It's something else.

sgeto:
Neither psutils nor imagemagick are needed for the pdf generation to work.
It's just something about the default groff executable.

If you remove psutils and imagemagick after updating groff it still works without them.

The groff command which does the job could be tested with cmake:
`groff -mandoc -Tpdf [man page]`

If it fails, we're dealing with a damaged groff executable.
Unfortunately the command needs a dummy file (a man page) as an argument or its exit code will always be 1. So testing with cmake is not that easy.

This issue is only present in Debian-based distros. This is a bug.
Personally, I think that it's neither ettercap's fault nor responsibility to make up for or catch this bug.
This PR is fine the way it is. If you want to wait until this issue is resolved by them (Debian), that's okay with me.

LocutusOfBorg:
I don't have such bug, anyway if you want to open one, I'll have a "Debian" look :)
(I didn't get the issue, and I can't reproduce on my machine)

koeppea:
When I'm back from vacation mid of next week, I'll try to check the CMake test. Or maybe @LocutusOfBorg finds some time since he's more settled down with CMake.
Generally I think we should not seeingly introduce caveats for some distros, even though it's not the fault of Ettercap.

LocutusOfBorg:
please also update README file :)

LocutusOfBorg:
@koeppea I can do it, but I honestly prefer to wait for a specific bug to fix :)
Do we need it? then cmake-check, otherwise we should understand and fix it in Debian

koeppea:
we don't need it yet. Can can as well continue living with ghostscript.
Who is going to file a bug report?

LocutusOfBorg:
I did:
create a pbuilder sid environment (minimal, 100Mb, just essential packages)
```
apt-get install groff --no-install-recommends
groff -mandoc -Tpdf *.in > log
```
no signs of errors.

koeppea:
You only installed *groff-base* package right?
Is then limited ot stable?

LocutusOfBorg:
ok, so we need groff and not only groff-base.
I wouldn't say this is a bug...

LocutusOfBorg:
that DESC file is in the binary groff and not groff-base
https://packages.debian.org/sid/amd64/groff/filelist

sgeto:
> ok, so we need groff and not only groff-base.

Yeah arguably not a bug. I think that when people say `groff` they mean "the real groff" not a limited basic version of it. 
The readme could be updated to make sure users install the full groff package or we could just stay with ghostscript.

**OR** there should be a cmake check to make sure groff understands the `-Tpdf` option.
I am fine with all three possibilities.

LocutusOfBorg:
```
$ apt-get install groff-base
$groff -Tpdf
groff: can't find `DESC' file
groff:fatal error: invalid device `pdf'
$ echo $?
3
$ apt-get install groff
groff -Tpdf < /dev/null
[foo]
echo $?
0
```

is this good enough? Anyhow, I don't see how much we will gain, maybe just a warning in the documentation is fine.
Running a command with cmake is easy to do
https://cmake.org/cmake/help/v3.0/command/execute_process.html

sgeto:
> is this good enough? Anyhow, I don't see how much we will gain, maybe just a warning in the documentation is fine.

I will improve this PR by adding a cmake check for -Tpdf soon
stay tuned :-D

sgeto:
I added a test for groff's -Tpdf option.
Sample fail output:
```
CMake Error at man/CMakeLists.txt:14 (message):
  groff (at /usr/bin/groff) does not seem to support PDF generation (3).

  groff failed with output:

  /usr/bin/groff: can't find `DESC' file

  /usr/bin/groff:fatal error: invalid device `pdf'



  Consider upgrading the groff installation.

  Users of Debian-based distributions must also install the full groff
  package.


-- Configuring incomplete, errors occurred!
```

@LocutusOfBorg & @koeppea can you please review this (and my two other PRs)?

koeppea:
Perfect. Thanks.

PR #840, opened 2017-10-11
Updated etter.finger.mac and improved get-mac.sh

sgeto:


koeppea:
wouldn't it even more generic using `$0` instead of the hard-coded script filename?

sgeto:
You mean [here](https://github.com/sgeto/ettercap/blob/753704c4c5cda9506b0e3310822c3759adca032f/misc/get-mac.sh#L4)?

koeppea:
yes correct.

sgeto:
Done.

LocutusOfBorg:
I would like to use the file as downloaded, not with changes useless.
We should change the code to use the standard oui.txt, and then use the system one when available. See:
https://packages.debian.org/sid/all/ieee-data/filelist
https://lintian.debian.org/tags/source-contains-data-from-ieee-data-oui-db.html

what do you think about?

koeppea:
Sorry I told completely bullshit.
*$0* will be `./get-mac.sh` when executed in the *misc* directory and `./misc/get-mac.sh` when executed in the *ettercap* directory and the test will always be true.

Can you please revert (reset) this change, squash your commits a little bit (the first 2) and rebase onto latest master?

sgeto:
@koeppea

I will once I get home...

sgeto:
> squash your commits a little bit (the first 2)

I don't know how to squash _just_ the first 2 commits. I just messed up my local repo while trying. :angry:
Is it really necessary?

koeppea:
it avoids cluttering the merge history
The [Wiki](https://github.com/Ettercap/ettercap/wiki/git-config) has some good advises helping to manage the local repository.

I you want, I can do that for you when you give me write access to your branch or fork.

sgeto:
You should have write access now.

koeppea:
@sgeto: yes you are right, your branch was quite messed up. I even couldn't find the commits there.
I exported the commits from the pull-request branch, reset yours, fast-forwarded it to the upstream/master and reapplied the exported commits and finally squashed the two get-mac.sh releated.

Should now be good.

@LocutusOfBorg 
> We should change the code to use the standard oui.txt, and then use the system one when available. 
> what do you think about?

I love that idea. You want to take on this? Otherwise I'll take care of it next week.
What about this pull request? Merge for now or hold back and see how well your approach goes down?

sgeto:
@koeppea thankx

koeppea:
@sgeto I don't understand why commit 4264018 repeats all the CMake changes that #872 should have already been done.

Why didn't you just do `git fetch upstream && git rebase upstream/master` to reapply the commits on top of current master?

sgeto:
Not sure really. I guess I wanted to keep the old "Updated etter.finger.mac" around. Or maybe I'm just too stupid. Funny thing is that it looks pretty ok from where I am standing.

I better just start over with a new PR from master before I break anything...

PR #843, opened 2017-10-21
Uncrustify second attempt

koeppea:
Since I've accidently closed the PR #516 from @LocutusOfBorg,
I've made up a new on based on his PR.
Of course if fine, I'm resetting the last commit, rebase over current master and reapply the uncrustify changes again.

I've adjusted some of the settings of the *uncrustify.cfg*.


sgeto:
I still think we should use clang-format instead of Uncrustify for the following reasons:

- it's more common. The most popular open-source projects use it.

- it's easy to configure.

- it doesn't introduce any additional dependencies. We already provide extensive support for clang/LLVM. This means every clang user already has clang-format installed. I'm also about to add clang-tidy support so these could go hand in hand.

@kholia is also in favor of clang-format. Is it possible to set up another poll?

eaescob:
How much effort will it be to switch to clang-format? I would hate for @koeppea 's work to go to waste

PR #855, opened 2017-12-12
fix non-gtk3 related warnings

koeppea:
This pull request addresses **openssl** and **GCC6** ( _-Wmisleading-indentation included in -Wall_) related warnings.
Some warnings cannot be fixed (yet):

1. **flex** (_at least in Debian_) is still experiencing a old bug: [#835542](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=835542)
2. **gtk3** related warnings cannot be easily fixed. Some could be, but wouldn't be backward compatible to GTK2. So I'm still working separating the GTK2 UI and redesigning the UI on the GNOME style GTK3 engine which more and more cannot be brought under the same hood.

### DIFF
--- a/src/dissectors/ec_ssh.c
+++ b/src/dissectors/ec_ssh.c
@@ -419,7 +419,8 @@ FUNC_DECODER(dissector_ssh)
                   return NULL;
                }
 #ifdef HAVE_OPAQUE_RSA_DSA_DH
-               RSA_get0_key(session_data->serverkey, &s_n, &s_e, &s_d);
+               RSA_get0_key(session_data->serverkey, 
+                     (const BIGNUM**)&s_n, (const BIGNUM**)&s_e, (const BIGNUM**)&s_d);
                get_bn(s_e, &ptr);
                get_bn(s_n, &ptr);
 #else
@@ -434,7 +435,8 @@ FUNC_DECODER(dissector_ssh)
                }
 
 #ifdef HAVE_OPAQUE_RSA_DSA_DH
-               RSA_get0_key(session_data->hostkey, &h_n, &h_e, &h_d);
+               RSA_get0_key(session_data->hostkey, 
+                     (const BIGNUM**)&h_n, (const BIGNUM**)&h_e, (const BIGNUM**)&h_d);
                get_bn(h_e, &ptr);
                get_bn(h_n, &ptr);
 #else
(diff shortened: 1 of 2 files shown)


sgeto:
There are tons of warning when building ettercap with GTK 2 (2.24.31) and 3 (3.22.26) support. BOATLOADS of warnings.
The scroll-back size of my terminal is currently set to 6000 lines and I am still not able to scroll back far enough to see the all warnings that compiling ec_gtk_view_connections.c and ec_gtk_view_profiles.c alone produce. It's that much... 
The majority of these warning are related to something being (explicitly mark as) deprecated. But there are other warnings as well.

koeppea:
As said. This pull explicitely omits GTK related warnings.
I'm working on this on a separate branch. The GTK3 code need to be separated from GTK2 since the compatibility diverged too far.

The GTK3 tree will undertake a major redesign of the GUI.

sgeto:
Where is the GTK3 tree? Is it public? I would like to check it out.
Major redesign you say...
Can we have a dark mode toggle in the GUI? THAT WOULD BE AWESOME!
And a toggle to stop/stop autoscrolling.
I could help with either. 
Especially with dark mode. :-D

What is your main motivation to keep gtk2 at all? Is it that some OSs that dont have gtk3?

koeppea:
So I'm aiming to redesign to GNOME like applications. If a dark scheme is used, ettercap's GUI should also fit this style.

I'm branch-wise I'm at the very beginning. Offline, I have something prepared that need to get integrated into the codebase.

The toggle of autoscroll (I assume you mean the message pane) is noted.
As soon as I have something worth a pull request, you're invited to contribute and improve.

The main motivation to keep GTK2 is mainly the "established" user community.
E.g. GNOME itself lost a huge user community after the switch to the new GUI w/o providing some kind of legacy style mode. This gap has meanwhile filled by MATE.

I'd like to have ettercap not to do the same mistake.

sgeto:
> If a dark scheme is used, ettercap's GUI should also fit this style.

Not sure I understand. I am referring to a toggle in the GUI to switch everything that is white to black. As far as I remember, the GNOME DE as a global switch for that too.

koeppea:
What about this pull request? Ready for merge?
@LocutusOfBorg @eaescob

LocutusOfBorg:
ok for me thanks!

PR #860, opened 2017-12-16
Don't create empty "desktop" folder if INSTALL_DESKTOP is set to off

sgeto:


eaescob:
👍

LocutusOfBorg:
ok for me, I would have changed
```
desktop/CMakeLists.txt
```

to remove the if(INSTALL_DESKTOP) check then

LocutusOfBorg:
```diff
--- a/desktop/CMakeLists.txt
+++ b/desktop/CMakeLists.txt
@@ -1,8 +1,8 @@
-if(INSTALL_DESKTOP AND ENABLE_GTK)
+if(ENABLE_GTK)
   configure_file(org.pkexec.ettercap.policy.in
     ${PKEXEC_INSTALL_WRAPPER}.policy @ONLY)
   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PKEXEC_INSTALL_WRAPPER}.policy DESTINATION ${POLKIT_DIR})
   install(FILES ettercap.desktop DESTINATION ${DESKTOP_DIR})
   install(FILES ettercap.appdata.xml DESTINATION ${APPDATA_DIR})
   install(PROGRAMS ettercap-pkexec DESTINATION ${INSTALL_BINDIR})
-endif(INSTALL_DESKTOP AND ENABLE_GTK)
+endif(ENABLE_GTK)

```
something like this, but a double check is probably better

koeppea:
@sgeto what do you think about @LocutusOfBorg suggestion?

sgeto:
Makes sense. I didn't look into `desktop/CMakeLists.txt` because I have no use for it. 
The redundant check in `desktop/CMakeLists.txt` should be removed. I'll write a new PR for this.

PR #870, opened 2018-02-20
New GTK3 UI

koeppea:
This pull request addresses originally the GTK related warnings when trying to build the GTK code against a current version of libgtk+3.x.

However, the way GTK libraries evolved made "fixing the current" code impossible, since major concepts have changed.
Therefore I decided to branch off a separate code directory for GTK3. This helps also to do a phased approach to decommission GTK2 support (if someday decided).

Given this constraint I took the chance and redesigned the UI a bit in order to keep up with modern GTK applications. This implicitly enables support of dark-themes since the UI integrates into GNOME or GTK based desktop environments.

To test, you need to set GTK_BUILD_TYPE to GTK3 when configuring using cmake.

Hope you like it.
ettercap-new-gtk3-ui-1
ettercap-new-gtk3-ui-4


sgeto:
Great job. Can't wait to test it.

gvanem:
I've not tested it. 

I just wish these patches could make the Gtk interface look as good as [HexChat](https://hexchat.github.io/]). I've never seen *any* Gtk based program on Windows that look so good as HexChat. I don't understand how it's possible without a heavily patched version of Gtk.  I gather it's due to [this Gtk](https://github.com/hexchat/gtk-win32) implementation. But that is *only* Gtk2. 

It seems your changes will not be Gtk2 compatible and work with the above Gtk2 code. Thus Ettercap (and other Gtk3-based apps on Windows) will still look rather ugly. 

Or do you have any idea how to make a Gtk3-based app on Windows look like HexChat? Just changing the`gtkrc` will not do this AFAICS.

koeppea:
You're right. This follows the path of GTK3 library. Maintaining GTK2 code depending on deprecated libraries isn't the way forward.

I've yet tested the UI on, Debian with MATE, GNOME and on Mac OS X. Surprisingly (after installing the icon theme) it looked perfectly the same as on a standard GNOME environment.

My next priory's is the fix-library. After that I'd have a look how HexChat is handling the UI on Windows. However would be still great if you could test this on Windows. GTK3 library is needed of course.

gvanem:
I fail to build this. It seems to require Gtk3.22 or newer. I have Gtk3.6.4 and there is no simple way for me to upgrade. The Gnome developers point us at some *Msysy2* and *pacman* crap. Besides:
https://stackoverflow.com/questions/32662215/how-do-you-install-gtk-3-0-on-windows

koeppea:
Thanks Gise,

Cmake should not build the gtk3 tree with a GTK version less than 3.12.

Does it is the same whan you try to build from master and set
GTK_BUILD_TYPE to GTK3? Just to see if it's a general thing we have to
look at in terms of Windows support or if it's spefic to the new GTK3 UI
introduced with this pull request..

gvanem:
I didn't use cmake for this test.

koeppea:
OK. Cmake normally takes care of the dependencies. As I said, it uses features introduced with 3.12. It won't build with a lower version. To cover that, the gtk tree is still responsible.

sgeto:
I think Hexchat uses gtk2. Their windows prebuilds does.
You can tell at first glance because gtk3 apps look very different in windows.
Unlike gtk2 apps (Hexchat, Zenmap and the old Wireshark UI), gtk3 apps don't follow the Windows UI at all. Not even in the title bar.
It looks very nice in my opinion, especially in dark mode. It's like Visual Studio and Adobe Photoshop had a baby.
I'll post a screenshot once I get to build this.
Maybe @koeppea can post one in the meantime?

@gvanem Msys2 is IMHO hands down a great project. It's very up-to-date. I use it a lot.
If you're looking for similar projects that provide up-to-date, MSVC-compatible gtk3 stuff, you should check out [gvsbuild ](https://github.com/wingtk/gvsbuild) and Microsoft's [vcpkg](https://github.com/Microsoft/vcpkg).
Under the hood, the former uses python and the later uses cmake 👅  But you should still checkout vcpkg. It's really good (it's a VS2017-only project).

koeppea:
> Maybe @koeppea can post one in the meantime?

done

sgeto:
Wow. Looks very sweet.
So we already talked about why gtk2 is still around. I get that.
But why is it the default?

koeppea:
> But why is it the default?

Good point. Let's have a discussion about that: https://doodle.com/poll/hittqmsdbszt2vm2

sgeto:
First build on Ubuntu 16.04 xenial errored. Using only system libs:
```
[ 61%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/gtk3/ec_gtk3_menus.c.o
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c: In function ‘gtkui_create_menu’:
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:108:50: error: ‘toggle_ip6scan’ undeclared (first use in this function)
       {"enable_ipv6scan",  NULL, NULL, DISABLED, toggle_ip6scan, {}},
                                                  ^
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:108:50: note: each undeclared identifier is reported only once for each function it appears in
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:108:7: warning: missing initializer for field ‘change_state’ of ‘GActionEntry {aka struct _GActionEntry}’ [-Wmissing-field-initializers]
       {"enable_ipv6scan",  NULL, NULL, DISABLED, toggle_ip6scan, {}},
       ^
In file included from /usr/include/glib-2.0/gio/gio.h:31:0,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:28,
                 from /usr/include/gtk-3.0/gdk/gdk.h:32,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3.h:4,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:23:
/usr/include/glib-2.0/gio/gactionmap.h:67:11: note: ‘change_state’ declared here
   void (* change_state) (GSimpleAction *action,
           ^
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:109:7: warning: missing initializer for field ‘padding’ of ‘GActionEntry {aka struct _GActionEntry}’ [-Wmissing-field-initializers]
       {"scan_hosts",  gtkui_scan, NULL, NULL, NULL, {}},
       ^
In file included from /usr/include/glib-2.0/gio/gio.h:31:0,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:28,
                 from /usr/include/gtk-3.0/gdk/gdk.h:32,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3.h:4,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:23:
/usr/include/glib-2.0/gio/gactionmap.h:72:9: note: ‘padding’ declared here
   gsize padding[3];
         ^
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:110:7: warning: missing initializer for field ‘padding’ of ‘GActionEntry {aka struct _GActionEntry}’ [-Wmissing-field-initializers]
       {"load_hosts",  gtkui_load_hosts, NULL, NULL, NULL, {}},
       ^
In file included from /usr/include/glib-2.0/gio/gio.h:31:0,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:28,
                 from /usr/include/gtk-3.0/gdk/gdk.h:32,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3.h:4,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:23:
/usr/include/glib-2.0/gio/gactionmap.h:72:9: note: ‘padding’ declared here
   gsize padding[3];
         ^
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:111:7: warning: missing initializer for field ‘padding’ of ‘GActionEntry {aka struct _GActionEntry}’ [-Wmissing-field-initializers]
       {"save_hosts",  gtkui_save_hosts, NULL, NULL, NULL, {}}
       ^
In file included from /usr/include/glib-2.0/gio/gio.h:31:0,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:28,
                 from /usr/include/gtk-3.0/gdk/gdk.h:32,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3.h:4,
                 from /mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_menus.c:23:
/usr/include/glib-2.0/gio/gactionmap.h:72:9: note: ‘padding’ declared here
   gsize padding[3];
         ^
src/interfaces/CMakeFiles/ec_interfaces.dir/build.make:1070: recipe for target 'src/interfaces/CMakeFiles/ec_interfaces.dir/gtk3/ec_gtk3_menus.c.o' failed
make[2]: *** [src/interfaces/CMakeFiles/ec_interfaces.dir/gtk3/ec_gtk3_menus.c.o] Error 1
CMakeFiles/Makefile2:342: recipe for target 'src/interfaces/CMakeFiles/ec_interfaces.dir/all' failed
make[1]: *** [src/interfaces/CMakeFiles/ec_interfaces.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

koeppea:
Interesting. Trying to reproduce (downloading ubuntu 16.04....)

sgeto:
Try without IPv6 support as well

koeppea:
64 or 32 bit?

> Try without IPv6 support as well

Does that mean that happens only when disabling IPv6 support?

sgeto:
64 bit.

> Does that mean that happens only when disabling IPv6 support?

I guessed that because of the error being `toggle_ip6scan` undeclared.
I just tried both settings:

Without IPv6 support, it errors at compiling `ec_gtk3_menus.c`
With IPv6 support, it errors at ettercap link time:
```
[ 65%] Linking C executable ettercap
interfaces/libec_interfaces.a(ec_gtk3.c.o): In function `gtkui_context_menu':
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3.c:1751: undefined reference to `gtk_menu_popup_at_pointer'
interfaces/libec_interfaces.a(ec_gtk3_plugins.c.o): In function `gtkui_plugin_context':
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_plugins.c:382: undefined reference to `gtk_menu_popup_at_pointer'
collect2: error: ld returned 1 exit status
src/CMakeFiles/ettercap.dir/build.make:146: recipe for target 'src/ettercap' failed
```
There are two warnings about the link error when conpiling (with and without IPv6):
```
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3.c: In function ‘gtkui_context_menu’:
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3.c:1751:8: warning: implicit declaration of function ‘gtk_menu_popup_at_pointer’ [-Wimplicit-function-declaration]
        gtk_menu_popup_at_pointer(GTK_MENU(data), (GdkEvent*)event);
        ^
```
and
```
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_plugins.c: In function ‘gtkui_plugin_context’:
/mnt/c/Users/Ali/projects/ettercap-master/src/interfaces/gtk3/ec_gtk3_plugins.c:382:7: warning: implicit declaration of function ‘gtk_menu_popup_at_pointer’ [-Wimplicit-function-declaration]
       gtk_menu_popup_at_pointer(GTK_MENU(menu), (GdkEvent*)event);
       ^
```

Maybe you forgot to conditionals.

koeppea:
Oh yes. Apparently gtk_menu_popup_at_pointer, which is the appropriate successor for gtk_menu_popup() has only introduced with GTK version 3.22 and the latter one deprecated with 3.22. 
So I'm adding an exception .... hope this is the only one.
BTW: Which GTK version have you installed?
```
$ pkg-config --modversion gtk+-3.0
```
The IPv6 action is now only being part of the array when build with IPv6 support.
Can you please recheck?

sgeto:
I recently upgraded to 3.18.9 on my Linux development box and I use 3.22.26 on Windows.
As much as I like to try 3.22.26, it's not that easy really. Keeping my windows patches up-to-date with the changes from this repo is a nightmare. Hope that changes soon.

sgeto:
There is one more `gtk_menu_popup_at_pointer()` in `src/interfaces/gtk3/ec_gtk3_plugins.c:382:7` that needs treatment...

src/interfaces/gtk3/ec_gtk3.c
```diff
@@ -1748,7 +1748,11 @@ gboolean gtkui_context_menu(GtkWidget *widget, GdkEventButton *event, gpointer d
    (void) widget;
 
     if(event->button == 3) {
+#if GTK_CHECK_VERSION(3,22,0)
```
sgeto: is this true for say gtk 3.22.26?

koeppea:
> There is one more

Thanks. Fixed. I rebased the last two commits and forced push. So you need to reset and pull again.

src/interfaces/gtk3/ec_gtk3.c
koeppea: > is this true for say gtk 3.22.26?

yes: https://developer.gnome.org/gtk3/stable/gtk3-Feature-Test-Macros.html#GTK-CHECK-VERSION:CAPS

sgeto:
Looks good now.

I guess that UI style (dark/light) is controlled globally via gnome's system settings or `gsettings`.
What about desktop environments that don't have either? How would they control this?
It would be nice if you could add a toggle for that. Not that would need it, but can get an idea for a switch from the gtk3-demo-application

screenshot 94

Possible?

cmake/Modules/EttercapLibCheck.cmake
```diff
@@ -35,18 +35,30 @@ if(ENABLE_GTK)
...
+        set(GTK3_FIND_VERSION 1)
         find_package(GTK3 REQUIRED)
         if(NOT GTK3_FOUND)
             message(FATAL_ERROR "You choose to build against GTK3, please install it, or build against GTK2")
         endif(NOT GTK3_FOUND)
-        set(HAVE_GTK3 1)
+        if (GTK3_MINOR_VERSION LESS 12)
+            message(STATUS "Your version of GTK3 (${GTK3_VERSION}) is not sufficient for full GTK3 support.")
+            message(STATUS "Full support requires >= 3.12. Building in GTK3 compatibility mode.")
+            set(GTK_BUILD_SUBTYPE "GTK3COMPAT")
```
sgeto: the GTK_BUILD_SUBTYPE variable seems unused...

sgeto: nevermind

gvanem: Did you try this Gtk3 on Windows too? If so, where/how did you get that dev-package?

sgeto: No. not yet.

Have you tried the two projects I recommended to you? [gvsbuild](https://github.com/wingtk/gvsbuild) and Microsoft's [vcpkg](https://github.com/Microsoft/vcpkg)

gvanem: These have no prebuilt .zip of Gtk > 3.12 AFAICS. Have to build yourself. Yuk!
Those Gnome nerds seems to loath Windows users. We'll be better of with wxWidgets and Qt5.

sgeto: > Have to build yourself. Yuk!

Yep. But it should be pretty straightforward. Especially with vcpkg.
The only prebuilds I know is the one that comes with msys2, and it's not msvc-compatiple.
I suppose you could also grab gtk3 prebuilds from appveyor's builder. You know, write a dummy yalm that copies zips and uploads them. 

>We'll be better of with wxWidgets and Qt5.

I have bad experiences with using Qt5. And the hell is it so big?

koeppea:
That is correct. Will check on the explicit option. But I think the theme need to support it somehow anyway. I’ll not go down the road and set colors in ettercap.

sgeto:
Fingers crossed.

koeppea:
Rebased over latest upstream/master.
Thanks @sgeto for helping me with the CMake files.

koeppea:
> Fingers crossed.

can you try the dark theme configuration option?

koeppea:
> can you try the dark theme configuration option?

Works for me. Tested on Debian running Fluxbox and in MacOSX

koeppea:
I bet Travis will fail for `$BUILD_ARGS = ""`.
Probably due to the introduction of **cmakelint** and the fact that the CMakeFiles in this branch may still violate.
@sgeto I've given you access to my _new-gtk3-ui_ branch. Could you help me with getting the CMake files compliant?

Moreover, there is no error message that the build failed due to the CMake formatting. Could you maybe address this in another pull request?

koeppea:
Got it. Now Travis happy again.

koeppea:
Now also tested on FreeBSD. Works there also perfectly except that the dark theme preference doesn't work (running XFCE DE). Installing now GNOME3 and retest.

koeppea:
OK apparently the GTK3 Theme installed with XFCE on FreeBSD didn't include a dark-theme. After installing GNOME, the configuration option has the effect as expected.

So now to me the only thing left is making GTK3 the default GTK_BUILD_TYPE.

sgeto:
I can't really test the option under my current setup without messing things up. If it's there and works on macOS it should work on Windoze as well.
If not, I'll hand in a complain then.

sgeto:
@gvanem
Hope I'm not too late:
Here is the smallest gtk3 (3.22.12) bundle I was able to create.
It's 32bit and without the icons.
If you use pkg-config, you **maybe** need to patch the .pc files in lib/pkgconfig.
I can build the full bundle if need one.
https://1drv.ms/u/s!AsbijvwHoSaBwWCZ_TJizDGeha0q

sgeto:
@koeppea

Have you taken a look at a way to prevent the output window from automatically scrolling?
There was no way to scroll back up without stopping the session because the UI kept jumping to end if a new message came up.

koeppea:
> the UI kept jumping to end if a new message came up.

Well this behaviour is there since from the beginning. This is controlled in the function `gtkui_msg()`. Nothing changed between the resident GTK2 and the new GTK3 code.

sgeto:
> Moreover, there is no error message that indicates that the build failed due to the CMake formatting. Could you maybe address this in another pull request?

Is this still a problem?

koeppea:
> Is this still a problem?

Not a problem. But kind of a insider. If only the build where `$BUILD_ARGS` are set to `""` fails, it's most probably a cmake syntax format issue.

I just had the idea to use a dedicated `$BUILD_ARGS` value, meaning nothing to cmake but just for travis that it shall include the cmake check there. Then it mayb be more obvious that the failure is due to the cmake format and a closer look at the output of **cmake-lint** may be benefitial.

sgeto:
Can you do it? Or would it be out of the scope of this PR?

koeppea:
Both. I could do it but I see it also outside of this PR.

sgeto:
> Not a problem. But kind of a insider. If only the build where $BUILD_ARGS are set to "" fails, it's most probably a cmake syntax format issue.

I get it. Would be nice if the entire build (all matrixes) would not even start if the test failed.
I don't know if travis has an option to set this. I will look for that. 
I guess I could add another variable to the first matrix to make it clear that cmake-lint is running there

> I could do it but I see it also outside of this PR.

Whatever you think is right. I just wanted to remind you.

koeppea:
yeah since I also don't know better about Travis' features,  lets start
with the way I proposed until a better idea / approach comes up.

I'll make up a new PR.

No Prob.

sgeto:
> I just had the idea to use a dedicated $BUILD_ARGS value, meaning nothing to cmake

Passing a dummy variable to cmake on the command-line (e.g "BUILD_ARGS =-DCMAKE_LINT=1") results in a warning message at the of the build configuration.
Adding a cmake independent environment variable introduces yet another set of matrixes...

sgeto:
never mind.
fixed here: https://travis-ci.org/sgeto/ettercap

gvanem:
@sgeto 
> It's 32bit and without the icons.

Thanks alot. I've installed it. But not tried to build with it yet. 
I could perhaps copy icons from my 3.6.4 version?

sgeto:
Merge?

sgeto:
I mean: Merge!

koeppea:
Let me rebase and squash some commits.

koeppea:
@sgeto when Travis builds the rebased commits fine, feel free to merge.

PR #868, opened 2018-02-14
Fix #793 - fix alignment of ettercap filters - update changelog

koeppea:
This fixes issue #793.
The main issue is, that the instructions are not (if not by accident) aligned on the address-bus width boundary.
Fixing the issue is two-fold:
 1. check byte-alignment when reading filters
 2. compiling filters with aligned instructions

Unfortunately this implicitly leads to invalidate (nearly) all pre-compiled filters. They need to get recompiled from the filter-source to get a aligned filter. But this shouldn't be a big deal, since **ettercap** checks the version of the filter anyway.
 
This PR aligns the instructions on 8-byte boundary even though on 32-bit systems, 4-byte alignment would be sufficient.
However in order to maintain filters compiled on a 32-bit system able to run on a 64-bit system I stick to 8-byte since it's automatically a 4-byte alignment.

PR #866, opened 2018-02-12
Fix issue #865 - Regex compilation prevents sslstrip plugin to load

koeppea:
This PR fixes an "invalid range end" regex issue.
In addition the error message is annotated with the string representation of the acutal error.

PR #872, opened 2018-02-27
Added a cmake linter

sgeto:
cmake-lint is a Python tool that checks and reports coding/style issues in CMake related build scripts.
Installation:
https://github.com/richq/cmake-lint
The recommended version is 1.4, which, at the time of this writing, was not yet published on PyPI.

From now on, ettercap uses cmake-lint (wrapped in a shell script) to enforce a consistent, portable and up-to-date cmake coding practice. This means that all cmake files **contributed** to the project will undergo a series of style/portability checks before the travis build phase can begin.

If this script finds any issues, it will exit with a non-zero status and a detailed error message. Else it won't output anything and exits with status code 0, continuing the build process.

The following rules are currently check for:

- Lines should be <= 120 characters long
Should be 80 lines but CMake versions prior to 3.0 do not support continuation with '\'.
- Expression repeated inside endif; use only endif()
...or only else(), endmacro(), endforeach(), endfunction() etc
- Do not use mixed case commands
It's all lowercase now
- Line ends in whitespace
- Extra spaces between 'some_function' and its ()
- Use (increments of) 2 spaces for indentation
- Use spaces for indentation not tabs
- Find modules should use uppercase names
So FindFOO.cmake not Findfoo.cmake
- Find Modules should include FindPackageHandleStandardArgs
Very useful. Also:
- Find Modules should use FIND_PACKAGE_HANDLE_STANDARD_ARGS
- Added script to report style/format issues in CMake files
- Updated .travis.yml to enforce these rules

The very first command run by travis reports format/style issues and exits appropriately.
This test will only happen once per commit and in first Matrix only. Not in every single Matrix.

With this PR, I have also already taken care of the existing issues, so we should only need to worry about staying clear of them in future commits. The integration with travis should help us do so.

TODO:
1. Enforce maximum line length of 80 characters (requires cmake version >= 3)
2. Fix (variables passed to std args in) FindGTK3.cmake
3. Move the description of this script to Etterwiki
4. Optional: Enable fast_finish in travis


koeppea:
I think this being a good idea especially in regard  to the main goal of the upcoming release.
However since the new-gtk3-ui as well as the fix-library branches wich are ought to be merged and doing massive changes in the CMake file, we should somehow coordinate the changes. Only one at a time.
So I'd like to define the order. But this is dependent on the estimation this will be ready?

sgeto:
I agree. It should be coordinated.

This PR's goal is to **not** change anything except fixing all existing issues and adding the script to control trigger travis.
It doesn't add or remove any feature.

There are two points left on my TODO list;
1. Enforce maximum line length of 80 characters (requires cmake version >= 3)
2. Fix (variables passed to std args in) FindGTK3.cmake  

These changes will require making changes to cmake. I want to do that seperatly.

So this is ready.
I think it's fine if this is merged after gtk but before opening a new PR for the fix-library branche.
You should go ahead and merge https://github.com/Ettercap/ettercap/pull/868 & https://github.com/Ettercap/ettercap/pull/866.
Having said that, this PR is meant to act as the basis for other cmake changes I have in mind. Which means that until this is merged, I won't be able to continue in this direction.
Maybe it's also a good idea to merge this before gtk. I can help you resolve any arising merge conflict.

koeppea:
> You should go ahead and merge #868 & #866.

ACK. Wait until these two are merged. Then rebase this PR over **upstream/master** and we can merge that.

> Maybe it's also a good idea to merge this before gtk.

ACK. I'll then rebase GTK branch over **upstream/master** and get back to you in case of any issues.
Then I'm doing the fix-library honoring the new CMake style rules.

> Enforce maximum line length of 80 characters (requires cmake version >= 3)

How to do this when you want to print a line but with CMake code indentation, you may end up in 20 characters left for the actual message? How to split a one-line message in two CMake code lines then?

sgeto:
Alright, I'll rebase this over upstream/master.

> How to split a one-line message in two CMake code lines then?

Right now it's not possible to escape newline characters in cmake files except within `message()`. Inside the `message()` function, you can even use  \n and \t to format your output. The feature to escape newlines everywhere else was introduced in cmake version 3. 

If I'm not mistaken, then Ubuntu 12.04 shipped with cmake 2.8.7 and Ubuntu 14.04 ships with cmake 3.2.2.  Ubuntu 12.04 is EOL so I am hoping to bump up the `cmake_minimum_required()` to 3 so we can escape newline characters.

sgeto:
At last. Ready to merge.

misc/cmakelint.sh
```diff
@@ -0,0 +1,57 @@
...
++readability/wonkycase,\
++syntax,\
++whitespace/eol,\
++whitespace/extra,\
++whitespace/indent,\
++whitespace/mismatch,\
++whitespace/newline,\
++whitespace/tabs
+
+CMAKELINT="cmakelint --filter=$FILTERS --quiet --spaces=2"
```
koeppea: is the setting of 2 spaces intended? Just asking because in our C-code style we expand tab to 3 spaces...

koeppea:
@eaescob good for me to go. I tested. Only formatting changes in _CMakeLists.txt_ files.
Cmake seem to work as before. When it's good for you, I'll go ahead an merge.

eaescob:
Looks good to me as well! Merging

misc/cmakelint.sh
sgeto: I tried a few settings and I personally think that (increments of) 2 spaces for indentation works best in cmake files.
The hard part is to find the setting in your editor to only do so in cmake files :-)

koeppea:
:+1:

Maybe appending a vim modeline...
Unfortunately the official documentation site is currently down.
I think it’s vim help „usr_21“.

sgeto:
I don't do vim. Never have.

koeppea:
But at least the ones that use it automatically profit from it. AFAIK some other editors also understand the syntax.
For others it’s a quick reference which syntax style is applied.

Von meinem iPhone

sgeto:
I guess.
Let me know if you get proper modeline config/thingy.

PR #873, opened 2018-02-27
Update CHANGELOG for upcoming release

koeppea:
This PR is one of the last PRs to be merged before releasing.
It captures the highlights and possible caveats of the new release.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,6 +16,7 @@ Legend:
    !! Lots of buffer under-/overflow conditions fixed
    !! CVE-2017-6430 (Fix invalid read on crafted file in etterfilter)
    !! fix dns_spoof plugin when used in bridge mode
+    + SSL redirects are now customizable at runtime
     + GeoIP detection / support using CMake
     + Rework of GTK3 UI - modern GNOME3 look
     + New Kerberos 5 downgrade plugin
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,29 @@ Legend:
          !! bug fixed
 
 =========================================
+0.8.3-?       2018mmdd
+   !! Non-aligned filters are no longer supported (recompilation with etterfilter required)
+   !! Fixed sslstrip plugin startup issue due to regex compilation error
+   !! Fixed lots of build warnings
+   !! Proper separation of library and executable code
+   !! Fixed heap-buffer-overflow in write_output in etterfilter
+   !! ip_addr sanity check when etterlog processes info logfile
+   !! Lots of buffer under-/overflow conditions fixed
+   !! CVE-2017-6430 (Fix invalid read on crafted file in etterfilter)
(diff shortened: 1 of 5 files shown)


koeppea:
Also copyright need to be updated

sgeto:
Do you mean the GPL?

koeppea:
No the copyright note contains the year.

sgeto:
Added to checklist

koeppea:
fa070d1: This what I found. Have I missed something else that is worth mentioning in the **CHANGELOG**?

sgeto:
Now we only need the name.

sgeto:
How do you intend to package the release? And under what compression/format?
If we use git, then we should make sure not to include such files as .travis.yml.

Another thing is that we can remove the need for bison and flex for people building from a release by adding the pre-processed files inside the tarball. I think libpcap does the same.
Thoughts?

koeppea:
> How do you intend to package the release? And under what compression/format?

The same way the past release packages.

> If we use git, then we should make sure not to include such files as .travis.yml.

True.


> Another thing is that we can remove the need for bison and flex for people building from a release by adding the pre-processed files inside the tarball.

AFAIK bison and flex are required for compiling filters. So I don't see that we don't need these as being dependencies even for "runtime".


> I think libpcap does the same.

I think I don't get it. We provide the source code and contributing files to build and install ettercap regardless of which packaging system is being used. The downstream maintainer takes care of that, depending on the Distribution. To build and Ettercap, the libpcap is necessary.

@sgeto Any idea when we could finish the Fix CMake pull request, so to complete the "code maintenance" release? I've already quite some new features in the pipe but don't want to mess with the  upcoming release.

koeppea:
Should we put #ettercap or #ettercap-project as IRC channel.
Myself is more hanging around in #ettercap-project.
What's your opinion?

eaescob:
we need to add ssl redirect rules now :)

CHANGELOG
```diff
@@ -5,6 +5,32 @@ Legend:
          !! bug fixed
 
 =========================================
+0.8.3-Belitto       2018mmdd
```
koeppea: Now when we know when to relase we can update the date which is then the commit to add the tag for 0.8.3.
I think the only thing left is to update the website.

koeppea:
I think everything is ready for releasing Monday.
 1. New release is drafted.
 2. Source-only tarball attached.
 3. Website is prepared.
   a) website distinguish between 3 different display sizes and optimized for mobile devices
   b) content has been updated
   c) just the new digest hashes are outstanding
 4. Screencast to present the new version is ready for upload to our YouTube channel.
 5. E-Mail to maintainers drafted.

Any objections / comments?

koeppea:
I'm still a bit unclear how ettercap.github.io and www.ettercap-project.org relate to each other.
@eaescob: Could you please explain? 
Thanks.

PR #875, opened 2018-03-01
Fix crashes when some external APIs fail

ZhouyangJia:
Hi,

I'm a PhD student. I analyzed the ettercap source code and found some potential API bugs that may cause crashes. These crashes are mainly caused by insufficient error handling of API functions like chdir or getsockopt.

I think it's unsafe to assume the library function would be correct. It would be better if we could handle the error properly.

Best,
Zhouyang

### DIFF
--- a/plug-ins/sslstrip/sslstrip.c
+++ b/plug-ins/sslstrip/sslstrip.c
@@ -771,7 +771,7 @@ static int http_get_peer(struct http_connection *connection)
    switch (ntohs(connection->ip[HTTP_CLIENT].addr_type)) {
       case AF_INET:
          if (getsockopt (connection->fd, SOL_IP, SO_ORIGINAL_DST, (struct sockaddr*)&ss, &ss_len) == -1) {
-            DEBUG_MSG("getsockopt failed: %s", strerror(errno));
+            WARN_MSG("getsockopt failed: %s", strerror(errno));
             return -E_INVALID;
          }
          sa4 = (struct sockaddr_in *)&ss;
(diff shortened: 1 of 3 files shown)


koeppea:
Good catch. Thanks.

src/interfaces/curses/widgets/wdg_file.c
```diff
@@ -123,7 +123,8 @@ static int wdg_file_destroy(struct wdg_object *wo)
    delwin(ww->win);
    
    /* restore the initial working directory */
-   chdir(ww->initpath);
+   if (chdir(ww->initpath))
```
koeppea: Why you don't check here explicitely against -1 but above?

src/interfaces/curses/widgets/wdg_file.c
```diff
@@ -123,7 +123,8 @@ static int wdg_file_destroy(struct wdg_object *wo)
...
+       perror("chdir failed");
```
koeppea: Why don't you use **DEBUG_MSG()** together with _strerror()_?

src/ec_sslwrap.c
```diff
@@ -816,13 +816,19 @@ static int sslw_get_peer(struct accepted_entry *ae)
 
    switch (ntohs(ae->ip[SSL_CLIENT].addr_type)) {
       case AF_INET:
-         getsockopt(ae->fd[SSL_CLIENT], SOL_IP, SO_ORIGINAL_DST, (struct sockaddr*)&ss, &ss_len);
+         if (getsockopt(ae->fd[SSL_CLIENT], SOL_IP, SO_ORIGINAL_DST, (struct sockaddr*)&ss, &ss_len) == -1) {
+            FATAL_ERROR("getsockopt failed");
```
koeppea: I'd rather use **WARN_MSG()** together with _strerror()_.

plug-ins/sslstrip/sslstrip.c
```diff
@@ -770,13 +770,19 @@ static int http_get_peer(struct http_connection *connection)
    socklen_t ss_len = sizeof(struct sockaddr_storage);
    switch (ntohs(connection->ip[HTTP_CLIENT].addr_type)) {
       case AF_INET:
-         getsockopt (connection->fd, SOL_IP, SO_ORIGINAL_DST, (struct sockaddr*)&ss, &ss_len);
+         if (getsockopt (connection->fd, SOL_IP, SO_ORIGINAL_DST, (struct sockaddr*)&ss, &ss_len) == -1) {
+            DEBUG_MSG("getsockopt failed: %s", strerror(errno));
```
koeppea: I'd rather use **WARN_MSG()**.

sgeto: I agree. Use WARN_MSG() instead.

sgeto:
👍

koeppea:
Merged Thanks.

PR #876, opened 2018-03-01
Fixed a compile warning

sgeto:
I think I just fixed the last compile warning.

🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

If so, how about compiling on travis with `-Werror` from now on?

### DIFF
--- a/plug-ins/gre_relay/gre_relay.c
+++ b/plug-ins/gre_relay/gre_relay.c
@@ -85,7 +85,10 @@ static int gre_relay_fini(void *);
 
 static void parse_gre(struct packet_object *po);
 static void parse_arp(struct packet_object *po);
+
+#ifdef WITH_IPV6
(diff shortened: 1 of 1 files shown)


koeppea:
Great.

I also spotted this one :-)

I don’t think we’d error on warnings. Some could sneak in from dependent libraries. We don’t have control over these. Current example when building on Debian: lex. I recall something like yy_parse or similar.

sgeto:
> Current example when building on Debian: lex. I recall something like yy_parse or similar.

I saw that too. Turns out that this a bug in flex. It was fixed in newer versions

koeppea:
Ok but my opinion is that we shouldn’t make our Travis builts relying on that.

sgeto:
fine

koeppea:
Merged thanks.

Issue #851, opened 2017-11-09
Etterlog regression


shark0der:
etterlog -p file.eci on master (ffb742e7a69bfc6fa71cac6bcd43cdf227eb2136) does not output any info. A much older version 9836548 works fine. Cannot provide more information at the moment since we have an ongoing conference :)

sgeto:
@shark0der is the conference finially over? 😄

shark0der:
Yes, it is :P
As said earlier, I ended up using 9836548 as master would just not output any info. AFAIK running strings on the eci file would show that the captured passwords (HTTP) are in the file but etterlog wasn't able to extract them. No errors, wasn't failing, just wouldn't extract those.

koeppea:
Could you please provide us with a test eci file so that we can try to reproduce?

koeppea:
OK I just created a test logfile. At least I agree that the current master doesn't print anything at all. However when I rewind back to the commit you mentioned I get a little bit more but ending in a segfault.
This may be the issue behind this change.
Let me have a deeper look.
Thanks for the heads up.

koeppea:
I think I found the issue.
However I'm only able to fix this along with the next major step to get the libettercap aligned. Otherwise this fix will just lead to yet another, but known issue.

Could you test my branch _fix-851_ with CMake option **ENABLE_GEOIP** set to **OFF**?

koeppea:
@shark0der : Sorry I forgot to commit and push the changes. So could you **now** please test my branch _fix-851_ with CMake option **ENABLE_GEOIP** set to **OFF**?

koeppea:
Closing as issue is fixed in #880

PR #698, opened 2015-08-14
Fix library

LocutusOfBorg:
Not yet ready :)

The solution might be:
- stop using the fake calls to printf for debug and use the libettercap version
  DONE: rename the gbls to some other thing, 
- use also the gbls from libettercap (the one you need BTW)
- create a "libettercap_init" with the proper inizialization code
- use "extern" variables to pass something to libettercap when we initialize (e.g. the pointers to something we use in the main)
- use libettercap code with no problems at all
- remove duplicated code (e.g. el_target.c) I tried to do it once if you remember, but I was stopped by the above points.

PR #880, opened 2018-03-12
Fix Library 2

koeppea:
This PR is the successor of #698.
I reworked the changes based on the commits from @LocutusOfBorg and fixed some missing pieces so that it finally builds fine with the use of shared library.
This way, the cross-referencing issues are gone and _etterlog_ and _etterfilter_ can make easy use of functions and macros in **libettercap**. 
All graphical stuff is handled by the shared library **libettercap-ui**, hence this code can also be shared among the different binaries.
This PR incorporates the fix of #851 which solves the issue now regardless if ettercap has been build with _GeoIP_ or not.
This PR is though not yet ready for merge. I first wan to see it building fine by _Travis_. Then some more changes will be committed that reduces Travis' exceptions and other Travis related changes.


sgeto:
You've been busy.

sgeto:
> ./cmake/Modules/EttercapLibCheck.cmake:14: Line ends in whitespace [whitespace/eol]

sgeto:
> /Users/travis/build/Ettercap/ettercap/src/interfaces/curses/widgets/wdg.h:157:4: error: unknown type name 'u_char'; did you mean 'char'?

koeppea:
> unknown type name 'u_char'; did you mean 'char'?

Haven't yet tried on MacOSX. There is one commit open from @LocutusOfBorg related to Mac.

koeppea:
> You've been busy.

A weekend visiting my extended family. I had a idea to **use** the time.

sgeto:
> I had an idea to use the time.

Very well spend. 👍 
Short type names are an issue on windows as well.
I think sys/types or the other ones should be included earlier.

koeppea:
I have the feeling it’s related to clang. If it behaves the same on FreeBSD it’s clear where to search. However haven’t had a chance to test this meanwhile.

src/ec_libettercap.c
```diff
@@ -0,0 +1,70 @@
...
+
+void libettercap_ui_start()
+{
+   /* start the actual user interface */
+   ui_start();
+}
+
+void libettercap_ui_cleanup()
+{
+   /* start the actual user interface */
```
sgeto: I saw this comment before somewhere... 🤔 
A few lines higher maybe?

koeppea: Will adjust the comment of course.

koeppea:
I think I’m close to getting behind the Mac building issue. Solving the u_char type issue is easy but the subsequent linking issue is somewhat evil. I’m pretty sure it’s related to the fact that on macOS, cmake builds the different parts a in a different order.

sgeto:
Fix library 2

Firstly:

So as far as Linux is concerned, this PR seems flawless:

- I think adding/changing a few CFLAGS/LDFAGS/CMake options would be goodto put my claim to the test. I'll prepare something shortly.
Btw what happened to asan support?   

- Not sure what to think of libettercap.[ch] with respect to its name. To me, it disrupts the established naming scheme.

- what do you think about adding a library version directly to libettercap?
I know it's mostly an internal API but especially for plugin developers, it would be useful to distinguish between the different APIs

---

Secondly:

Did you find the macOS problem? If so, you can mostly ignore the rest of this message.

> I’m pretty sure it’s related to the fact that on macOS, cmake builds the different parts a in a different order.

- I unfortunately don't have a Mac to test on. Do you know anything about running macOS inside a VM? I remember someone referring to this as "Hackingtosh".

- If you think the issue is the build order, you could try building it one target at a time.
Running "make help" should show you all available targets. From there you could then run "make libettercap", "make ettercap" and so forth.
I am assuming here that the dependancies' of each target in question aren't themselves mixed up to the extend that only editing the build scripts could fix this. Let me know what you think and which make targets you suspect to be causing the issue.

- I have the far-fetched theory that the issue may be the include order rather than the build order. That's because I saw something similar when I was experimenting with clang on windows and explicitly setting include directories fix it for me. Interested to hear your thoughts.

- Just a thought and a bit unrelated: One of libpcap'a developers, is quite experienced with the ins and outs of macOS and in a conversation he mentioned that Apple got rid of OpenSSL and/or is pushing they "own" implementation of it.
As far as I can tell it's available out of the box and is, for the most part compatible with OpenSSL.
We just need to tell CMake to look for and link with it instead. I can get the snippet for that from libpcap/master (I wrote part it). That would remove the extra brew related build steps.

koeppea:
> Btw what happened to asan support?

See #804 



> - Not sure what to think of libettercap.[ch] with respect to its name. To me, it disrupts the established naming scheme.

I don't get that.



> - what do you think about adding a library version directly to libettercap?

That's a good idea in my opinion. Will include this as well.



> - I unfortunately don't have a Mac to test on. Do you know anything about running macOS inside a VM? I remember someone referring to this as "Hackingtosh".

No I've organized myself a test machine.



> From there you could then run "make libettercap", "make ettercap" and so forth.

This is a very good hint. Since gdb debugging on Sierra really got very difficult

koeppea:
It's strange.
Since *lib_ettercap* provides the most commonly used exported functions, it need to be build first, which equals the build order on a non-MacOS system: e.g. beginning with `ec_asn1.c`.
When I try to explicitly build *lib_ettercap* first, it builds the interfaces first and when it tries to link it together to the *libettercap-ui* library, linking fails, since the exported functions from *lib_ettercap* are not yet build. At least this is my understanding:
```
$ make lib_ettercap
Scanning dependencies of target ec_interfaces
[  0%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/__/ec_interfaces.c.o
[  1%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text.c.o
[  1%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text_conn.c.o
...
[ 27%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/gtk3/ec_gtk3_view_profiles.c.o
[ 29%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/gtk3/ec_gtk3_shortcuts.c.o
[ 29%] Linking C shared library libettercap-ui.dylib
Undefined symbols for architecture x86_64:
  "_add_ip_list", referenced from:
      _curses_host_target1 in ec_curses_hosts.c.o
      _curses_host_target2 in ec_curses_hosts.c.o
      _add_target1 in ec_curses_targets.c.o
      _add_target2 in ec_curses_targets.c.o
      _gtkui_button_callback in ec_gtk3_hosts.c.o
      _add_target1 in ec_gtk3_targets.c.o
      _add_target2 in ec_gtk3_targets.c.o
      ...
...
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[3]: *** [src/interfaces/libettercap-ui.dylib] Error 1
make[2]: *** [src/interfaces/CMakeFiles/ec_interfaces.dir/all] Error 2
make[1]: *** [src/CMakeFiles/lib_ettercap.dir/rule] Error 2
make: *** [lib_ettercap] Error 2
```
Can we maybe meet some time late Friday in IRC. We could do a Teamviewer session on my Mac so that we could troubleshoot this together. Let me know if that works out for you.
Maybe you have an idea in the meantime, why it tries to build interfaces first.

sgeto:
Sorry today doesn't work for me. The whole weekend in fact. Let's arrange a new date via mail.


> Since lib_ettercap provides the most commonly used exported functions, it need to be build first


I think you're mixed something up. As far as I can tell, the UI lib has always been build first (actually, if enabled, ec_lua is build first, then ettercap-UI, then libettercap...). 
The UI lib also shouldn't depend on anything exported by libettercap. It's actually the other way around: 
At the very minimum, libettercap needs select_deamon_interface() and select_text_interface() (found in ec_interfaces.c, which is part of the UI lib) and that's why ettercap-UI must be build first.
See also target_link_ library(lib_ettercap ...) in src/CMakeList.txt


Honestly the macOS port will eventually build and run. I'm pretty confident that we'll figure it out before this PR is ready for merger.

Just keep going with whatever else you have planned for Fix Library 2. It will make things clearer and us wiser.
If you want, you can add the failing build matrixes to allowed_failures in .travis.yml and remove them before merger (e.i. after they're fixed).

koeppea:
> Sorry today doesn't work for me. The whole weekend in fact. Let's arrange a new date via mail.

alright



> I think you're mixed something up. As far as I can tell, the UI lib has always been build first

Yes you're right. I've seen this also as I tried the same on Linux the same evening...

> f you want, you can add the failing build matrixes to allowed_failures in .travis.yml and remove them before merger (e.i. after they're fixed).

macOS is a support OS. So for me, this is not a option. Maybe to postone the shared library on that platform and use static instead for macOS. But I don't give up yet. 
I'll dig down the extra warnings from the patch you provided to me.

sgeto:
> Maybe to postone the shared library on that platform and use static instead for macOS

Postpone, yeah sure. That's what I meant.
But using/linking libettercap statically isn't really an option. I wish it was though. Has it ever worked for you? Linking with a libettercap.a results in a mess on my machine, where every dissector becomes unavailable.

koeppea:
Yeah I found it finally.
The symbol lookup could be fixed by a special linker flag dedicated to the MacOSX linker: `-undefined dynamic_lookup`.

But then it was becoming tricky. Starting **ettercap** right away resulted in a SEGFAULT.
I knew it was due to accessing `ec_gbls` in `main()` but I haven't seen it for weeks..... 😒 
Today I found the glitch in the code 💡 
`ec_gbls` was declared twice: in _ec_main.c_ and in _ec_globals.c_ .
Now since both source files are linked in different binaries/library it matters: in **libettercap** `ec_gbls` is properly allocated but back in **ettercap**, the local declaration masks `ec_gbls` from **libettercap** and results in dereferencing NULL pointer.
This seem to be special to the MacOSX compiler and/or linker, since this issue didn't occur on any other OS.

sgeto:
:+1 :+1 :+1
At last!

koeppea:
@sgeto @LocutusOfBorg @eaescob This one is ready in my opinion. Objections?
If not, please merge so that @sgeto can rebase then master over his pull request #891 and finish that.

koeppea:
rebased over current master and resolved conflict.

PR #892, opened 2018-04-21
fix interface combo by selecting the name but display the description

koeppea:
Hi @sgeto, @gvanem 

this pull request should fix the startup issue of the GTK3 UI on Windows.
Effectively, it just fixes the handling of the interface name and description properly.
It also adds a debug message of which interface _name_ has been selected.
Could you please check and give feedback?

Cheers Alex


sgeto:
You did it again!! Thanks.
Works as expected. 

screenshot 101

(My interface descriptions are shorter in the screenshot than in the log I send you. That's because I also tested this against my fork of wpcap. It uses the adapters "friendly names" instead of the npcap's buggy  interface descriptions)

koeppea:
But wait. Let me check who much effort it is to add the IP also. ...

koeppea:
I think it's not such a good idea. It would probably destroy the geometry of the setup screen.
ready for merge. Should also not interfear #880.

PR #893, opened 2018-04-24
Fix format string bug in etterlog --filter

droberson:
A format string bug exists within etterlog:

$ etterlog --filter %p

etterlog 0.8.2 copyright 2001-2015 Ettercap Development Team

TARGET (0x5557eca62b6a) contains invalid chars !

Looking at the code, I saw two ui_fatal_error() functions. One utilized va_args, the other didn't. I did not see any usages of the va_args-enabled ui_fatal_error() function elsewhere in the code.

fatal_error() within ec_error.c calls ui_fatal_error() without format strings, triggering this bug.

### DIFF
--- a/utils/etterlog/el_ec_compat.c
+++ b/utils/etterlog/el_ec_compat.c
@@ -57,16 +57,6 @@ void ui_error(const char *fmt, ...)
    fprintf(stderr, "\n");
 }
 
-void ui_fatal_error(const char *fmt, ...)
-{
-   va_list ap;
-   va_start(ap, fmt);
-   vfprintf (stderr, fmt, ap);
-   va_end(ap);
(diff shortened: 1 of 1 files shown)


koeppea:
Thanks but sorry. This is a duplicate. The upcoming pull-request #880 already addresses and eliminates these issues.
It's already working for Linux and *BSD. We've to fix issues on MacOSX to get it ready.

sgeto:
Hey,

> etterlog 0.8.2 copyright 2001-2015 

Should probably say "etterlog 0.8.2 copyright 2001-2017".
Try this on current master and see if you get the same results. I don't. 

"We are saddened by a bird's cry, but not for a fish's blood.
Blessed are those with voices."

-Mamoru Oshii

PR #921, opened 2019-01-19
Allow NDP and ARP poisoning without any optional parameters

primeos:
According to the man page both parameters for the NDP MITM attack are
optional:
```
ndp ([remote],[oneway])
```

But ettercap aborted with the following error message:
```
$ sudo ettercap -Tq6i eth0 -M ndp /// ///
[...]
FATAL: NDP poisoning: missing arguments.
$ echo $?
255
```

With this patch the NDP MITM attack without any parameters works as
expected.

### DIFF
--- a/src/interfaces/curses/ec_curses_mitm.c
+++ b/src/interfaces/curses/ec_curses_mitm.c
@@ -60,11 +60,12 @@ struct wdg_menu menu_mitm[] = { {"Mitm",                'M', "", NULL},
 static void curses_arp_poisoning(void)
 {
    char *method = "arp:";
+   char *default_param = "remote";
    size_t len = strlen(method);
    
    DEBUG_MSG("curses_arp_poisoning");
 
-   strncpy(params, method, len);
+   snprintf(params, PARAMS_LEN, "%s%s", method, default_param);
 
    curses_input("Parameters :", params + len, PARAMS_LEN - len - 1, curses_start_mitm);
 }
@@ -109,11 +110,12 @@ static void curses_dhcp_spoofing(void)
 static void curses_ndp_poisoning(void)
 {
    char *method = "ndp:";
+   char *default_param = "remote";
    size_t len = strlen(method);
 
    DEBUG_MSG("curses_ndp_poisoning");
 
-   strncpy(params, method, len);
+   snprintf(params, PARAMS_LEN, "%s%s", method, default_param);
 
    curses_input("Parameters :", params + len, PARAMS_LEN - len - 1, curses_start_mitm);
 }
--- a/src/interfaces/gtk/ec_gtk_mitm.c
+++ b/src/interfaces/gtk/ec_gtk_mitm.c
@@ -40,7 +40,6 @@ void gtkui_arp_poisoning(void)
    GtkWidget *dialog, *vbox, *hbox, *image, *button1, *button2, *frame, *content_area;
    gint response = 0;
    gboolean remote = FALSE;
-   gboolean oneway = FALSE;
 
    DEBUG_MSG("gtk_arp_poisoning");
 //   not needed, the \0 is already appended from snprintf
@@ -90,22 +89,17 @@ void gtkui_arp_poisoning(void)
       const char *s_remote = "", *comma = "", *s_oneway = "";
 
       if(gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON (button1))) {
-	s_remote="remote";
+        s_remote="remote";
         remote = TRUE;
       }
 
       if(gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON (button2))) {
          if(remote)
-		comma = ",";
+            comma = ",";
 
-	 s_oneway = "oneway";	
-         oneway = TRUE;
+         s_oneway = "oneway";
       } 
 
-      if(!remote && !oneway) {
-         ui_error("You must select at least one ARP mode");
-         return;
-      }
       snprintf(params, PARAMS_LEN+1, "arp:%s%s%s", s_remote, comma, s_oneway);
       gtkui_start_mitm();
    }
@@ -254,7 +248,7 @@ void gtkui_port_stealing(void)
       if(gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON (button2))) {
          if(remote)
               comma = ",";
-	 tree = "tree";
+         tree = "tree";
       }
   
       snprintf(params, PARAMS_LEN+1, "port:%s%s%s", s_remote, comma, tree); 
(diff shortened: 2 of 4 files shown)


koeppea:
Looks good for me. :+1:

koeppea:
Tested in GTK GUI. NDP poisoner doesn't start and throws error message.
How have you tested it? In text-mode?

primeos:
@koeppea Sorry, I've only tested it with the text only interface and didn't know that there is an extra check for the GUI.

I've removed that check for both GTK2 and GTK3 and also noticed that there is a check for ARP so I removed that as well (`-M arp: /// ///` already works with the text only interface).

Note: I didn't test the curses interface as I couldn't get it to work properly (after starting unified sniffing (`eth0`) and scanning for hosts the menus disappear and I don't get a message that any hosts where found).

koeppea:
Alright I’m going to test this weekend and feed back since I’m off Keyboard for the next few days

sgeto:
> 
> 
> Looks good for me. 👍

Same here.

koeppea:
Withou having it checked. But have you also checked Curses UI also?

primeos:
No, as I noted above:

> Note: I didn't test the curses interface as I couldn't get it to work properly (after starting unified sniffing (eth0) and scanning for hosts the menus disappear and I don't get a message that any hosts where found).

I've just rebuild everything from master and this is what happens in my Parrot VM:

screenshot_parrot_2019-01-25_12 06 17
screenshot_parrot_2019-01-25_12 06 43

Can you reproduce that?

koeppea:
So first building produces now a warning:
```
warning: variable ‘oneway’ set but not used [-Wunused-but-set-variable]
    gboolean oneway = FALSE;
             ^~~~~~
```
I've tried but couldn't reproduce:
curses

primeos:
> So first building produces now a warning:

Sorry, I've forgot to build with `Debug` and missed that. It's fixed now and I've also replaced a few tabs in the affected files with spaces for consistency.

primeos:
> I've tried but couldn't reproduce:

Interesting, not sure on what this could depend (probably `ncurses`, build configuration, hardware differences, terminal emulator, etc.). Unfortunately I cannot investigate this ATM (bit short on time :o).

primeos:
> If you give me write permissions to your repo I can take care for the curses part for you.

Ah, forgot to mention, I looked through the source-code and didn't find any checks like for the GTK UI. I've just verified it on another VM (Ubuntu LTS) to be sure, but feel free to double check it.

> I'll also consolidating the micro commits to one.

Already did that for you in case it helps and I'll give you write access in case you also want to change something else ;)

koeppea:
Yes your're right. Curses UI doesn't have an extra check / error message.

primeos:
> Only connections not leaving the LAN are in this mode subject to ettercap's dissection.
> So what is your intention?

For my use case it actually doesn't matter. It's for a local test network without a gateway so the result will be the same. But since I'm using the text interface I didn't see the need to specify `remote` (which worked with ARP but after the switch to IPv6 it didn't work with NDP as expected). Since this behaviour (exit with an error message) was unexpected and I assume there are a few use cases for this (only local attacks) I thought I'd make a PR.

> I'm thinking to predefine the "remote" as I think, this is what most users would expect. Of course predefining doesn't apply for the TEST UI.

+1, I agree and guess it wouldn't hurt (as long as it remains optional) :)
Not sure about the last part though, you mean the text UI, right (i.e. change the default for the Curses, GTK2, and GTK3 GUI)?

Should I change it or will you change it?

primeos:
The `curses` part isn't as compact as I hoped it would be but it seemed to fit best into the existing codebase. Is that approach ok or would you prefer something else?

src/interfaces/curses/ec_curses_mitm.c
```diff
@@ -60,13 +60,17 @@ struct wdg_menu menu_mitm[] = { {"Mitm",                'M', "", NULL},
 static void curses_arp_poisoning(void)
 {
    char *method = "arp:";
-   size_t len = strlen(method);
+   size_t len1 = strlen(method);
+
+   char *default_param = "remote";
+   size_t len2 = strlen(default_param);
```
koeppea: `method` and `default_param` are both initialized on the stack then copied into another stack buffer concatenated.
I don't see the benefit having the param and method in two separate variables and copy twice.
Either use `snprintf()` to perform the copy in one shot or I'd prefer, having one stack variable per MITM method (ARP|NDP) and copy it once into the global `params` buffer.

src/interfaces/curses/ec_curses_mitm.c
```diff
@@ -109,13 +113,17 @@ static void curses_dhcp_spoofing(void)
 static void curses_ndp_poisoning(void)
 {
    char *method = "ndp:";
-   size_t len = strlen(method);
+   size_t len1 = strlen(method);
+
+   char *default_param = "remote";
+   size_t len2 = strlen(default_param);
```
koeppea: Same applies here.

src/interfaces/curses/ec_curses_mitm.c
```diff
@@ -60,13 +60,17 @@ struct wdg_menu menu_mitm[] = { {"Mitm",                'M', "", NULL},
...
    
    DEBUG_MSG("curses_arp_poisoning");
 
-   strncpy(params, method, len);
+   strncpy(params, method, len1);
+   strncpy(params+len1, default_param, len2);
 
-   curses_input("Parameters :", params + len, PARAMS_LEN - len - 1, curses_start_mitm);
+   curses_input("Parameters :", params + len1, PARAMS_LEN - len1 - len2 - 1, curses_start_mitm);
```
primeos: The problem is basically this part. I need `len1` and `len2` to avoid showing `arp:` in the input field for the parameters and correctly limit the length. I'm not that familiar with C and didn't really have a better idea. One possibility would be to use `params + 4*sizeof(char)` instead of `params + len` which would work. Would that solution be ok? The drawback is a hard-coded constant (`4` for the characters "arp:") but in this case it probably won't be a problem as it should become obvious during testing (and "arp:" will probably never change anyway).

koeppea:
Let me think of it....

koeppea:
Would you mind giving me temp. write access to your repo. I'd make a suggestion.

koeppea:
What do you think of that approach?

primeos:
LGTM, way cleaner, thanks :)

I've also just noticed (since you've improved that as well) that I actually didn't have to subtract `len2` here:

```
curses_input("Parameters :", params + len1, PARAMS_LEN - len1 - len2 - 1, curses_start_mitm);
```

koeppea:
When you like it you can squash the commits together and force push

primeos:
Ok, did that. Thanks for the help :)

koeppea:
Merged. Thanks.

PR #926, opened 2019-02-06
Fix new warnings due to new gcc version

koeppea:
Since GCC7, some things trigger warnings which wasn't the case before.
Fixing the code avoiding these warnings while retaining the functionality.
Additionally the intendiation has been harmonized in `ec_inet.c`.

### DIFF
--- a/src/dissectors/ec_bgp.c
+++ b/src/dissectors/ec_bgp.c
@@ -179,7 +179,7 @@ FUNC_DECODER(dissector_bgp)
          
          /* Get authentication data */
          if (length > 1) {
-            snprintf(PACKET->DISSECTOR.pass, 4, "Hex(");
+            snprintf(PACKET->DISSECTOR.pass, 5, "Hex(");
             str_ptr = PACKET->DISSECTOR.pass + strlen(PACKET->DISSECTOR.pass);
 
             for (j = 0; j < (length-1); j++) {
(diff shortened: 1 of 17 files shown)


koeppea:
Hey @LocutusOfBorg @sgeto .
Have you had a chance to check this pull?

koeppea:
@LocutusOfBorg @sgeto would be nice if you could check this. I've a bigger pull-request in preparation which I'd like to rebase over the recent fixed.

LocutusOfBorg:
Hello, it looks sane to me, I get why they are needed, to be sure the termination char is copied too, right?

but I have a general question: is that extra char something we send over the network? this might break some protocols, or maybe we can just think wrt initializing the data with "0", but I guess making gcc silent is of course the best thing to do!

koeppea:
Hi @LocutusOfBorg thanks for having a look.
Yes most changes are due to the 0 byte.
From `strncpy()`'s man page:

>        The strncpy() function is similar, except that  at  most  n  bytes  of  src  are
>        copied.   Warning:  If there is no null byte among the first n bytes of src, the
>        string placed in dest will not be null-terminated.

From `snprintf()`'s man page:

>        The functions snprintf() and vsnprintf() write at most size bytes (including the
>        terminating null byte ('\0')) to str.

There are a view more minor things.

I'll double check on the protocol breaking. If this is the case, the the **mem*** - functions would better fit such cases. If I don't see the risk, I'll then go ahead and merge if that OK for everybody.

Thanks.

koeppea:
I couldn't spot anything that would break a protocol. So will merge now.

PR #935, opened 2019-04-06
Customizable SSL redirect rules

koeppea:
This pull request adds a feature that is very helpful when used with SSL intercept.
In general, if SSL intercept is enabled, traffic for all interceptable protocols is being redirected through ettercap.

This has the drawback, that for any target site, security warnings may be thrown or even worse, the user can't make it to the destination at all, because the target is protected by HSTS.

In case, that not all SSL traffic is of interest but just for a specific destination, the redirection should customizable.

Of course, one could adjust the redirect rules on OS level while ettercap is running, which is quite cumbersome. From a user-friendly view, it's just all or nothing.

This pull-request's goal is to change this. Ettercap starts with it's all-or-nothing ruleset. But now the user can adjust this ruleset even before MITM is started, to the interessting SSL traffic in question using the selected user interface.

**Example from a FreeBSD machine**
image

_Resulting ruleset_
image


LocutusOfBorg:
I really like it! I didn't test it, but looks ok to do...
Do you plan also to update some manpage for it? or maybe it isn't needed, I'm not sure at all...

koeppea:
Good point.... 

Von meinem iPhone gesendet

koeppea:
@LocutusOfBorg this way? I'll make another Youtube Video to introduce the new UI and this feature.

eaescob:
Oh wow, this is really useful! I haven't tested yet, but how are results so far?

koeppea:
Well I have it tested on Debian and freebsd.
Well it’s just about ettercap to control the local redirect rules. No other behavior like ARP spoof or SSL interception is changed.
But would be good if someone else could test it too.

eaescob:
The code makes sense to me. I don't have the env set up for testing. Let's merge and allow others to submit issues, if any. When are we due for a new version?

koeppea:
> Am 18.06.2019 um 04:06 schrieb Emilio A. Escobar <notifications@github.com>:
> 
>  When are we due for a new version?
> 

From my point of view we should go for releasing. Last release is already long time ago and we’ve pushed many bugfixes and enhancements in the meantime.

eaescob:
Agreed. Let's get the changelog updated and go for release
Emilio Escobar    / eescobar@gmail.com

koeppea:
For code name: what do you think of Bertillon?

eaescob:
I like it or Locard

koeppea:
Locard was already used for 0.7.6

eaescob:
That's why it looked so familiar. Let's go with yours!

koeppea:
I'll prepare the web site over the weekend so that could either release on 24.06. or if we wait for one more week we can schedule the release for 01.07.

sgeto:
Bertillon as in "*Alphonse Bertillon*"?

eaescob:
That's the one

Emilio Escobar    / eescobar@gmail.com

sgeto:
Exciting.

PR #939, opened 2019-04-24
Fix #937 - Build failure in ec_stdint.h/wdg.c on Alpine with GCC 8.3

### LINKED ISSUE #937: Build failure in ec_stdint.h/wdg.c on Alpine with GCC 8.3
I have been attempting to build ettercap at commit ffeaac725d31f10b270cd3ece14e5a7b83a1f1ff on Alpine but have experienced the following issue:

```
# make
Scanning dependencies of target ec_interfaces
[  1%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/__/ec_interfaces.c.o
[  1%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text.c.o
[  1%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text_conn.c.o
[  2%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text_display.c.o
[  2%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text_plugin.c.o
[  3%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text_profile.c.o
[  3%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/daemon/ec_daemon.c.o
[  3%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/curses/widgets/wdg.c.o
In file included from /ettercap/e/src/interfaces/curses/widgets/wdg.h:21,
                 from /ettercap/e/src/interfaces/curses/widgets/wdg.c:22:
/ettercap/e/include/ec_stdint.h:37:10: error: unknown type name 'u_int8_t'
  typedef u_int8_t   u_int8;
          ^~~~~~~~
<snip>

In file included from /ettercap/e/src/interfaces/curses/widgets/wdg.c:22:
/ettercap/e/src/interfaces/curses/widgets/wdg.h:161:4: error: unknown type name 'u_char'
    u_char screen_color;

<output truncated>
```

I do not experience the same issue on Ubuntu 18.04.

However, upon further investigation, it appears that neither `ec_stdint.h` or `wdg.h` explicitly import `sys/types.h` for Linux (`wdg.h` does for Darwin) which is where these types are defined. 

On Ubuntu 18.04, it appears that `sys/types.h` is being implicitly imported within `wdg.h` when it imports `stdlib.h`. I.e.

```bash
$ echo "#include <stdlib.h>" | gcc -E - | grep -e "\(u_int8_t\)\|\(u_char\)"
typedef unsigned char __u_char;
typedef __u_char u_char;
typedef unsigned int u_int8_t __attribute__ ((__mode__ (__QI__)));
```

However, on Alpine this is not the case. I did not trace the imports to see the difference.

It appears there are two possible solutions, the first is to tweak `ec_stdint.h` to unconditionally include `sys/types.h` on Linux (perhaps for other OS'es as well?). That is, by adding:

```C
#if defined(OS_LINUX)
        #include <sys/types.h>
#endif
```

Alternatively, `wdg.h` can be updated as follows:
```diff
--- a/src/interfaces/curses/widgets/wdg.h
+++ b/src/interfaces/curses/widgets/wdg.h
@@ -14,7 +14,7 @@
    #include <windows.h>
 #endif
 
-#ifdef OS_DARWIN
+#if defined(OS_DARWIN) || defined(OS_LINUX)
    #include <sys/types.h>
 #endif
```

I am inclined to think the first is the better solution, however, the second appears to be more aligned with the existing code base. If needed, I can submit a PR.

kazkansouh:
Minor tweak to imports in `wdg.h` to unconditionally include `sys/types.h` on Linux.

Fixes #937

### DIFF
--- a/src/interfaces/curses/widgets/wdg.h
+++ b/src/interfaces/curses/widgets/wdg.h
@@ -14,7 +14,7 @@
    #include <windows.h>
 #endif
 
-#ifdef OS_DARWIN
+#if defined(OS_DARWIN) || defined(OS_LINUX)
    #include <sys/types.h>
 #endif
 


sgeto:
I think if we do this we have to do it right. These old type names are all over the source and making up for them one issue/PR at a time is just inefficient.
This PR even compensates for it instead of upgrading them. All our dependancies use the new type, even libnet. I don't see why we shouldn't. Thoughts?

koeppea:
@sgeto What do you mean exactly?
I'm afraid I don't get it.

sgeto:
I mean that we should completely switch to the c99 data types defined in stdint.h instead of making up for issues we are and will be facing by continuing to use these deprecated pre-std u_*_t types.

koeppea:
Yes that make sense.
@kazkansouh Could have a look at #941

koeppea:
Closing in favor of #941

Issue #942, opened 2019-05-05
Ettercap filter explicit value


luismartinsez:
I am trying to create an ettercap filter to establish a MITM attack in Modbus TCP traffic.

Could help me to understand what parameters I should be passing in the following "Search" and "Replace"  for changing register 1?

if (ip.proto == TCP && tcp.dst == 502) {
   if (search(DATA.data, "\xE1\x00")) {
    msg("ModbusValue Register 1 Changed from 225 to 3000");
      replace("\xE1\00", "\xBB\x80");

![image](https://user-images.githubusercontent.com/48541300/57187834-a5f1b680-6f27-11e9-987a-1311e68b6c35.png)

When I send the above script, it changes values of all Registers from 1 through to 12, but I need to change the value of only Register 1. Is it possible explicitly define the script to change only Register 1?

luismartinsez:
There's an x missing in previously posted script. Following is the correct one.

if (ip.proto == TCP && tcp.dst == 502) {
if (search(DATA.data, "\xE1\x00")) {
msg("ModbusValue Register 1 Changed from 225 to 3000");
replace("\xE1\x00", "\xBB\x80");

koeppea:
So these two bytes just identify the value, but not the register.

I assume there are a few bytes before that identify the register.

You need to include this in your seach/replace pattern, of course when
replacing, you've to let the bytes to identify register 1 unmodified.

However I'm not aware of the internals of Modbus.

koeppea:
Can you attached (zipped) the packet capture file?

luismartinsez:
Thank you for your response.

I am following a paper published by Gabriel Sanchez of SANS Institute, which is comprehensive. However, the researcher uses a filter to make changes to a single coil register value. In my case, I am trying to change the value of a holding register.

Uploaded pcap & Gabriel’s paper.

ModbusPal can be used to analyse data structure and traffic of Modbus TCP protocol. https://github.com/zeelos/ModbusPal

luismartinsez:
[koeppea.zip](https://github.com/Ettercap/ettercap/files/3146296/koeppea.zip)

koeppea:
So I'm wondering why it even replaced anything.
Because you're only looking at the payload if the TCP destination port is 502.
But actually, the desired content only appears in the response where the TCP port 502 is the source - not the destination.

However, I don't know if this is due to missing curly brackets - but I hardly can't imagine etterfilter would have compiled it.
[...]

PR #945, opened 2019-05-24
etterfilter: fix early null-byte termination

koeppea:
This PR intends to fix a bug in **etterfilter** reported by #942.

The use of `strdup()` in combination with binary string matching and assignment terminates too early in case a null-byte is in the middle of the to-be assigned or to-be match string: e.g. `\xab\x00\xcd`.
This either assigns random data following the zero-byte or compares a shortened byte sequence.

This pull request replaces the use of `strdup()` with a combination of `calloc()` and `memcpy()`.
First tests show that it's fixing the issue reported in #942.

Testing with the reporter of #942 outstanding.


eaescob:
This makes sense to me, thanks!

LocutusOfBorg:
Did we forget lots of SAFE_FREE there? @koeppea

PR #948, opened 2019-06-20
Depreciation warning if building against GTK2

koeppea:
Warning during CMake configuration to make more clear about the GTK2 phase-out.


eaescob:
👍

PR #951, opened 2019-07-04
Update version to 0.8.3 in CMakeLists.txt

LocutusOfBorg:
This should be needed to avoid libettercap being versioned as 0.8.2


LocutusOfBorg:
thanks!

PR #965, opened 2019-09-07
include user message output in debug logfile

koeppea:
This pull requests aims to improve the troubleshooting by including the output on the messages pane via `USER_MSG()` into the debug logfile if Ettercap has been build in Debug mode.

This way it's easier to put the rest of the logfile into the context what actually happened visible for the user.


eaescob:
This makes sense to me

Issue #962, opened 2019-08-22
Etter.dns issues (dns_spoof)


himuura:
Can't figure it out why but when arp poisoning and dns_spoofing a target a couple of things happen:

- if the line (example) www.microsoft.com PTR is commented or does not exist, the spoof doesnt work, i cannot catch a single DNS request.
- if the refered line is changed accordingly, i can see the dns_spoof in ettercap but on the victim's browser i only see connection reseted. All internet connection is suspended, i mean i cannot access to a single web page (like a DoS).

I've enabled ip forward, uncommented the lines for iptables on etter.conf, set the ID's to 0. The columns on the etter.dns file are tab separated or space? like microsoft.com (tab) A (tab) IP ?

Any ideas on how to solve this one? Im testing both Kali linux latest and raspbian buster latest, and i'm experiencing the same behaviour on either one. Apache is running and i can access the demo page from other devices.

koeppea:
Please read this [wiki page](https://github.com/Ettercap/ettercap/wiki/Providing-debug-information)

koeppea:
Please reproduce with your version of _etter.dns_.

himuura:
Weird, after following the wiki page, ettercap -G breaks almost automatically...

[     init]	ssl_wrap_init
[     init]	sslw - bind 992 on 59263
[     init]	ec_redirect("insert", "telnets", IPv4, 0.0.0.0/0, 0.0.0.0/0, 992, 59263)
[     init]	ec_redirect(): execute [iptables -t nat -A PREROUTING -i eth0 -p tcp -s 0.0.0.0/0 -d 0.0.0.0/0 --dport 992 -j REDIRECT --to-port 59263]
[     init]	register_redir_service(telnets)
[     init]	ec_redirect("insert", "telnets", IPv6, ::/0, ::/0, 992, 59263)
[     init]	 !!! SEGMENTATION FAULT !!!
[     init]	ui_cleanup

Any idea?

himuura:
To where can i send you the .pcap log?

koeppea:
pack it in a tarball and attach it to the issue

himuura:
[dnsspoof.tar.gz](https://github.com/Ettercap/ettercap/files/3540329/dnsspoof.tar.gz)

To sum up the PCAP:

[...]

PR #968, opened 2019-09-12
Autostart CLI provided plugins on graphical UI

koeppea:
This pull-request is inspired by issue #962 to align the behavior when plugins are provided on the CLI using the  `-P` parameter.

Prior this pull-request only the text interface acts on provided plugins on the CLI.
The command like `ettercap -G -Pdns_spoof` does effectively nothing with the provided plugin and it remains deactivated.

With this pull request, all user interfaces treat the provided plugins and start them as soon as Ettercap's initial setup is succeeded.

This also improves troubleshooting plugin related problems since activation of the plugin is no longer depended on the user's interaction on the graphical user interface.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,13 @@ Legend:
          !! bug fixed
 
 =========================================
+0.8.4-XXXXXXXXX    00000000
+    + CLI provided plugins are now also autostarted in graphical UI
+
+
+
+
+
 0.8.3-Bertillon    20190701
    !! Fix binary comparsion and assignment in etterfilter
    !! Fixed packetbuffer racecond. in BRIDGE mode (e.g. Message too long)
--- a/man/ettercap.8.in
+++ b/man/ettercap.8.in
@@ -714,6 +714,12 @@ More detailed info about plugins and about how to write your own are found in
 the man page ettercap_plugins(8)
 
 .TP
+\fB\-\-plugin-list <PLUGIN1>[,<PLUGIN2>,...]\fR
+Instead of providing multiple occurances of \-P plugin, \-\-plugin-list can be
+used followed by a comma sepaparated list without any spaces.
+(e.g. ./ettercap \-\-plugin-list plugin1,plugin2).
+
+.TP
 \fB\-F\fR, \fB\-\-filter <FILE>\fR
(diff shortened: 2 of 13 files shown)


himuura:
how do i pull grom this?

koeppea:
This is described [here](https://github.com/Ettercap/ettercap/wiki/git-config).

himuura:
Done mate thanks! Check the issue with dns_spoof whenever you can. Cheers!

sgeto:
I love the idea.

As an aside:
I remember that it took me a while to learn how to enable multiple plugins at the same time. The correct syntax is 
```
ettercap [...] -Pplugin1 -Pplugin2 -Pplugin3 ...
```
This is also undocumented. So maybe we could add it to the docs or allow enabling plugins a bit more elegantly. E.g:
```
ettercap [...] -Pplugin1, plugin2, plugin3 ...
```
Or maybe even without the comma. But then, some plugins takes/need additional parameters. So if we do this we should probably stick with the comma after all...

koeppea:
Your'r right @sgeto, will add documentation. However, I'm afraid your idea breaks parsing the options as space introduces the next option or argument. To get this done will become quite ugly and very uncommon.
Potentially using comma but w/o spaces, but I'm not sure if it's worth the effort.

sgeto:
> Potentially using comma but w/o spaces, but I'm not sure if it's worth the effort.

Using comma-separated lists is how we're adding lua scripts and arguments. You know, those how let me 😄 

From `src/ec_parser.c`:
```
--lua-script <script1>,[<script2>,...]     comma-separted list of LUA scripts\n"
--lua-args n1=v1,[n2=v2,...]               comma-separated arguments to LUA script(s)\n"
```
I understand your concern for the spaces but getting this to work with commas should be a matter of copy-pasting those code snippets.

koeppea:
@sgeto, like this?

eaescob:
I really like this change, thanks! Looks good to me, @himuura @sgeto  any other feedback?

himuura:
yup, -P is now working as intended. Cannot, at times, capture a dns_spoof but i think it may be related to the fact that i am wireless connected. gave it a couple of tests and seemed to work fine with the -P flag.

Issue #966, opened 2019-09-09
ettercap does not detect harfbuzz include path


jelly:
Pango now depends on /usr/include/harfbuzz/hb.h which breaks the build of ettercap on Arch. Some changes in the cmake file are required.

```
Scanning dependencies of target ec_interfaces
[  0%] Building C object src/interfaces/CMakeFiles/ec_interfaces.dir/__/ec_interfaces.c.o
In file included from /usr/include/pango-1.0/pango/pango-font.h:25,
                 from /usr/include/pango-1.0/pango/pango-attributes.h:25,
                 from /usr/include/pango-1.0/pango/pango.h:25,
                 from /usr/include/gtk-3.0/gdk/gdktypes.h:35,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:30,
                 from /usr/include/gtk-3.0/gdk/gdk.h:32,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from /build/ettercap/src/ettercap-0.8.3-gtk/src/interfaces/gtk3/ec_gtk3.h:4,
                 from /build/ettercap/src/ettercap-0.8.3-gtk/src/ec_interfaces.c:31:
/usr/include/pango-1.0/pango/pango-coverage.h:28:10: fatal error: hb.h: No such file or directory
   28 | #include <hb.h>
      |          ^~~~~~
compilation terminated.
```

Maybe this can be seen as an inspiration.

https://cgit.kde.org/kde-gtk-config.git/commit/cmake/modules/FindGTK3.cmake?id=cf8db4ebe5ef5e54517604cd9330535983e4dff0

koeppea:
Thanks. This will definitely help.

koeppea:
Same issue as in #957. Travis build is currently also failing due to this changed dependency. Pull request #969 is supposed to fix that but not yet finished.

koeppea:
@jelly can you please try if #969 fixes that for you?

koeppea:
@jelly any update? Have you had a chance to try the proposed fix for the harfbuzz issue?

koeppea:
PR #969 should solve the build issue. Closing.

PR #969, opened 2019-09-13
Fix travis build

koeppea:
Several changed dependencies need to be treated to make Travis happy again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ addons:
       - libbsd-dev
       - libcurl4-openssl-dev
       - libgtk2.0-dev
+      - libidn11-dev
+      - librtmp-dev
       - libltdl-dev
       - libluajit-5.1-dev
       - libncurses5-dev
@@ -138,9 +140,10 @@ matrix:
       env: BUILD_ARGS="-DENABLE_GTK=Off -DENABLE_CURSES=Off"
 
 before_install:
-    - if [ ${TRAVIS_OS_NAME} = "osx" ]; then brew update; brew install check curl; brew install groff gtk+ gtk+3; brew install libidn libnet; brew install luajit openssl pcre rtmpdump geoip; fi
+    - if [ ${TRAVIS_OS_NAME} = "osx" ]; then brew update; brew install check curl; brew install groff gtk+ gtk+3 harfbuzz; brew install libidn libnet; brew install luajit openssl pcre rtmpdump geoip; fi
 before_script:
     - if [ ${TRAVIS_OS_NAME} = "osx" ]; then export BUILD_ARGS="$BUILD_ARGS -DOPENSSL_ROOT_DIR=`brew --prefix openssl`";fi
+    - if [ ${TRAVIS_OS_NAME} = "osx" ]; then export BUILD_ARGS="$BUILD_ARGS -DHARFBUZZ_DIR=`brew --prefix harfbuzz`"; fi
 
 script:
   - if [[ "$CC" == "gcc" && "${BUILD_ARGS}" = "" ]]; then git clone https://github.com/richq/cmake-lint.git; cd cmake-lint; python setup.py install --user; cd -; ./misc/cmakelint.sh; fi
--- a/cmake/Modules/EttercapLibCheck.cmake
+++ b/cmake/Modules/EttercapLibCheck.cmake
@@ -89,6 +89,15 @@ Please install it, or build against GTK1")
         if(FOUND_GTHREAD)
             set(EC_INTERFACES_LIBS ${EC_INTERFACES_LIBS} ${FOUND_GTHREAD})
         endif()
+        #
+        # harfbuzz is dependent by pango but is not found by
+        # FindGTK3.cmake on MacOS X.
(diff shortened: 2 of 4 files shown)


sgeto:
Can you try my modified `FindGTK3.cmake` module?
[FindGTK3.zip](https://github.com/Ettercap/ettercap/files/3618645/FindGTK3.zip)

The HarfBuzz stuff is integrated into FindGTK3, so make sure to remove the entries in `EttercapLibCheck.cmake` - or not, whatever works best I guess. 😅 

I think that the `FindHarfBuzz.cmake` module you added should already be part (newer) cmake versions. So maybe you can check if you already have that file and leave the local copy out.
See [commit f0c9e72e](https://gitlab.kitware.com/cmake/cmake/commit/f0c9e72ee152e1e4acda8efeed9c2e3e408b2658#58185a7528d74478217625e6c21fe5e83e2f9302)

koeppea:
Thanks @sgeto, will do.

koeppea:
Alright. Done. Let see if Travis becomes happy.

koeppea:
Hi @sgeto,

good news, the _FindGTK3.cmake_ you've provided detects successfully harbuzz on Linux.
bad new, it isn't also not able to find Harfbuzz on MacOSX --> Travis failures.

Do you know where this cmake file originated from? Maybe we can query the author.

sgeto:
> Do you know where this cmake file originated from? Maybe we can query the author.

Sorry no. Last time I checked `FindGTK3.cmake` was passed around a few times and then abandoned.

Correct me if I'm wrong: The reason cmake is unable to find the header is because it's in a unusual directory, right? According to #966 `hb.h` is  in `$PREFIX/usr/include/harfbuzz/hb.h`?

Maybe we can bring our local `FindHarfBuzz.cmake` back and tweak `find_path()` to additionally look in that directory?

[FindHARFBUZZ.zip](https://github.com/Ettercap/ettercap/files/3634399/FindHARFBUZZ.zip)

Can you try this?

sgeto:
I think [this line](https://travis-ci.org/Ettercap/ettercap/jobs/587967471#L2257) means that cmake found hb.h but not the library.
Another thing we could try is to set the environment variable `HARFBUZZ_DIR` instead of `HARFBUZZ_ROOT_DIR` in `travis.yml`. The find module seems to somewhat honor this variable.

I going to assume now that the library is located somewhere in `/usr/lib` and attach an updated 
`FindHarfBuzz.cmake` that is setup to check there as well. Fingers crossed!

[FindHARFBUZZ.zip](https://github.com/Ettercap/ettercap/files/3639450/FindHARFBUZZ.zip)

sgeto:
Whoa, frustrating...
I got more:
I found this module that uses the pkgconfig to locate harfbuzz.
[FindHarfBuzz.zip](https://github.com/Ettercap/ettercap/files/3639896/FindHarfBuzz.zip)

koeppea:
Seems I've to grab a real Mac to better troubleshoot....

sgeto:
Good idea.

koeppea:
Well your last version of the file seem to work.
Just CMake Linter is failing. Just confirmed the style and then Travis should be happy again.
Before mergin I'll of course fixup all the commits.

koeppea:
Should be good to merge now. @sgeto What do you think?

sgeto:
I think the stuff in `EttercapLibCheck.cmake` is redundant. We should leave it out.
I am also not very happy with having introduce yet another dependency (pkg-config) to all of ettercap just because of macos. Pkg-config was only needed when building the tests, meaning that people didn't need to have to build ettercap.
Personally, I think the fact that cmake's own `FindHARFBUZZ.cmake` module is unable to locate harfbuzz is a bug and that should be addressed upstream. I guess it will be at some point question only is when.
If possible it would be nice if we could fall back to pkg-config stuff or limit it to macos only.

koeppea:
I'd don't think thats a too big issue, since this is only relevant if the OS is MAC.
However I don't think that the Stuff in `EttercapLibCheck.cmake` can't be omitted since `FindGTK3.cmake` alone cannot find it, at least not on MAC. However I can try if you want.

Regarding CMake upstream: I've already opened a issue for the CMake Folks for the same issue, but referencing to `FindGTK3.cmake`. They said that they do not provide that. So I'm pretty sure `FindHARFBUZZ.cmake` is also not officially from CMake.

koeppea:
@sgeto Any further comment? Are we good to merge?
This way I'll be able to rebase the other pending pull requests.

koeppea:
Hey @sgeto have you had a chance to look over this PR?

koeppea:
Hi @sgeto You've been busy last weeks isn't it? Could you please do the review of this PR as it's blocking all remaining PRs from getting Travis build successful.
Thanks.

koeppea:
Since @sgeto seems to be unresponsive ATM, I hope you @LocutusOfBorg can take over reviewing this PR.

sgeto:
Sorry. I am going through a rough time. This PR is of course okay.
Again sorry for the huge delay.

You may merge it now. Or I'll if GitHub won't let you...

koeppea:
Thanks Ali

PR #971, opened 2019-10-02
Fix TLS handshake issue

koeppea:
This pull request aims to fix TLS handshake failures when SSL is intercepted with Ettercap.
  1. Firstly the issue with new TLS versions, that the default private key in `etter.ssl.crt` isn't long enough for ECDH calculations. 
  2. The SAN extension is now taken over from the server certificate
  3. The SNI from the client-side ClientHello is now taken over in the server-side ClientHello (required OpenSSL 1.1.1)

Still some browsers deny a intercepted connection (seen on Ubuntu w/ Firefox resulting in SEC_ERROR_BAD_SIGNATURE) which seems to be, an issue on the browser side with extra security tools enabled. Chrome and curl (even with TLS1.3 enforced) works.


eaescob:
Seeing build error: libettercap.so.0.8.3: undefined reference to `TLS_server_method'
libettercap.so.0.8.3: undefined reference to `TLS_client_method'

Do we need to specify OpenSSL version?

koeppea:
I’ll have a closer look.

Von meinem iPhone gesendet

koeppea:
It's because Travis has a OpenSSL version < 1.0.2g. Apparently the `TLS_client_method` and `TLS_server_method` functions have been there only in later versions. Built in another OpenSSL version check to mask the symbols to their predecessor names if required.

PR #976, opened 2019-11-19
Fix #975 - don't wait for keypress when easter egg is triggered

### LINKED ISSUE #975: "Birthday" easter egg breaks scripts
I am running this command in my script: `ettercap -T -s 's(300)olqq'` (from the man page)

Due to the need to press the enter key, I have to send the enter key via stdin in my script, else ettercap just hangs forever.

This obviously is easily worked around but it could've been a problem if I wasn't debugging the script this very day.

koeppea:
Fix #975 
This pull request just omits the waiting to press any key to continue starting ettercap when the easter egg has triggered.
@eaescob: I put you as reviewer to decide whether or not you want this behavior changed.

### DIFF
--- a/src/ec_main.c
+++ b/src/ec_main.c
@@ -51,8 +51,6 @@
 
 /* protos */
 
-static void time_check(void);
-
 /*******************************************/
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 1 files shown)


sgeto:
Looks good.
I can't decide whether the easter egg should be there at all. I mean it's kinda cool but then, their email addresses are dead...

Feel free to merge.

eaescob:
I see the build failed on travis but has nothing to do with this. I'd say we can get rid of the easter egg overall.But this is the safest change, okay to merge

koeppea:
Now I updated the PR to remove the easter egg at all.

Issue #974, opened 2019-11-10
Ettercap :0 hosts added to the hosts list...


Achraf-ai:
Hello , i was working with ettercap but whereever i scan my network ettercap says : 0 hosts added to the hosts list... even when i want to do arp spoofing in Terminal : FATAL: ARP poisoning needs a non empty hosts list.
i have already 8 machine in my network 
I work with kali linux 2019 not in virtual machine in real machine

ettercap -T -q -M ARP /192.168.1.1// /192.168.1.10// output:

ettercap 0.8.2 copyright 2001-2015 Ettercap Development Team

Listening on:
 wlan0 -> 02:BD:51:71:7B:53
	  192.168.1.17/255.255.255.0
	  fe80::28b6:bf2a:8293:77d7/64

Privileges dropped to EUID 0 EGID 0...

  33 plugins
  42 protocol dissectors
  57 ports monitored
20388 mac vendor fingerprint
1766 tcp OS fingerprint
2182 known services
Lua: no scripts were specified, not starting up!

Scanning for merged targets (2 hosts)...

* |==================================================>| 100.00 %

0 hosts added to the hosts list...

FATAL: ARP poisoning needs a non empty hosts list.

koeppea:
I see you're using wireless. Can you try on wired network?
Also you're not using the latest version.
Please read [our Wiki](https://github.com/Ettercap/ettercap/wiki/Providing-debug-information) to help us troubleshooting your issue.

Pixel48:
Same problem in eth0 with 0.8.2

koeppea:
Please note that 0.8.2 is not the latest code release from Github.
What about the wired test?

tijldeneut:
This problem is persistent, "arping" command works perfectly. Both VM/native, wired or wireless, Kali up-to-date …
Both GUI as Text based version of Ettercap has the same issue, I've build 0.8.3 and it remains.
I assume some change in ethernet stack in newest Kali?

tijldeneut:
To confirm: Ettercap on Kali 2019.3 (possibly after an apt dist-upgrade) is broken
# apt dist-upgrade -y
With built-in Ettercap 0.8.2 and again with a freshly compiled 0.8.3, I opened Wireshark: the ARP requests for the entire subnet are leaving and ARP replies are sent back, exactly as expected. 
But the Host List for Ettercap GUI remains empty and the error message with the Text version remains:
FATAL: ARP poisoning needs a non empty hosts list.

[...]

PR #977, opened 2019-11-24
Fix PCAP

koeppea:
Starting with **libpcap** version 1.9.1, some changes come in that affect Ettercap:
   1. `pcap_lookupdevice()` to determine the _default_ interface has become deprecated
   2. Setting `PCAP_TIMEOUT` to 0 on Linux (used with `pcap_open_live()`) causes a huge delay between the call of `pcap_loop()` and starting  to pass packets to the handler function `ec_decode()`. This has the effect, that the network scan results in a 0 host list because the scan wait's only 1 second until it cancels the capture thread.

This PR addresses both issues. Of course, behavior changes on other platforms still need to be confirmed. However this depends on the availability of the new **libpcap** version on these platforms.

### DIFF
--- a/include/ec_capture.h
+++ b/include/ec_capture.h
@@ -11,6 +11,7 @@ EC_API_EXTERN EC_THREAD_FUNC(capture);
 
 EC_API_EXTERN int is_pcap_file(char *file, char *pcap_errbuf);
 EC_API_EXTERN void capture_getifs(void);
+EC_API_EXTERN char* capture_default_if(void);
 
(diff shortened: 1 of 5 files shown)


koeppea:
This PR fixes #974

eaescob:
Should we include pcap 1.9.1 with ettercap? Or specifically ask for it?

koeppea:
I think with this PR we're fine wither either version. So I see no need to provide it as bundled.

PR #988, opened 2019-12-21
Update cmake FindGTK3 module to cope with new pango-1.0 harfbuff requ…

LocutusOfBorg:
…irement, fixing this kind of build failures:

See:
https://gitlab.gnome.org/GNOME/pango/issues/387
and
https://github.com/eiskaltdcpp/eiskaltdcpp/commit/8b22479aef21fb2b764051f0499ae5fba397d31f
For background information

/<<PKGBUILDDIR>>/src/interfaces/text/ec_text.c: In function ‘text_redirects’:
/<<PKGBUILDDIR>>/src/interfaces/text/ec_text.c:715:4: warning: ignoring return value of ‘fgets’, declared with attribute warn_unused_result [-Wunused-result]
  715 |    fgets(input, 20, stdin);
      |    ^~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/pango-1.0/pango/pango-font.h:25,
                 from /usr/include/pango-1.0/pango/pango-attributes.h:25,
                 from /usr/include/pango-1.0/pango/pango.h:25,
                 from /usr/include/gtk-3.0/gdk/gdktypes.h:35,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:30,
                 from /usr/include/gtk-3.0/gdk/gdk.h:32,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from /<<PKGBUILDDIR>>/src/interfaces/gtk3/ec_gtk3.h:4,
                 from /<<PKGBUILDDIR>>/src/ec_interfaces.c:31:
/usr/include/pango-1.0/pango/pango-coverage.h:28:10: fatal error: hb.h: No such file or directory
   28 | #include <hb.h>
      |          ^~~~~~
compilation terminated.
make[4]: *** [src/interfaces/CMakeFiles/ec_interfaces.dir/build.make:66: src/interfaces/CMakeFiles/ec_interfaces.dir/__/ec_interfaces.c.o] Error 1
make[4]: *** Waiting for unfinished jobs....
[  2%] Creating PDF for /<<PKGBUILDDIR>>/obj-graphical/man/ettercap_plugins.8


koeppea:
I thought we've had addressed the Harfbuzz issue with #969 
On which platform these build failures occur?

LocutusOfBorg:
They happens on Ubuntu focal or Debian experimental, they both have new pango

koeppea:
OK the question then comes up if the harfbuzz changes from #969 (dedicated CMake file) is still required or can be omitted.

LocutusOfBorg:
I did the change, lets see if travis is good with that!

LocutusOfBorg:
travis looks good, btw we don't have any failures left...

LocutusOfBorg:
it seems good now!
we might want to simplify it a lot by getting rid of gtk2

koeppea:
Looks promising. Testing on all supported OS'es...

koeppea:
Currently on FreeBSD, _make_ is failing with:
```
Scanning dependencies of target ettercap
[ 66%] Building C object src/CMakeFiles/ettercap.dir/ec_parser.c.o
[ 66%] Building C object src/CMakeFiles/ettercap.dir/ec_main.c.o
[ 67%] Linking C executable ettercap
ld: error: libettercap.so.0.8.3: undefined reference to SSLeay_version
ld: error: libettercap.so.0.8.3: undefined reference to SSL_library_init
ld: error: libettercap.so.0.8.3: undefined reference to SSLv23_client_method
ld: error: libettercap.so.0.8.3: undefined reference to SSLv23_server_method
ld: error: libettercap.so.0.8.3: undefined reference to X509_set_notAfter
ld: error: libettercap.so.0.8.3: undefined reference to X509_set_notBefore
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1
```
investigating....

koeppea:
I tried to isolate the issue. It happens as soon as `interfaces/libettercap-ui.so.0.8.3` is added to the linker. w/o linking against `interfaces/libettercap-ui.so.0.8.3`, it's building w/o error.
Further investigating....

.travis.yml
```diff
@@ -112,10 +113,10 @@ matrix:
       env: BUILD_ARGS="-DENABLE_PDF_DOCS=On -DENABLE_IPV6=On -DENABLE_LUA=On -DCMAKE_BUILD_TYPE=Debug"
     - os: osx
       compiler: gcc
-      env: BUILD_ARGS="-DENABLE_PDF_DOCS=On -DENABLE_IPV6=On -DENABLE_LUA=On -DGTK_BUILD_TYPE=GTK3"
+      env: BUILD_ARGS="-DENABLE_PDF_DOCS=On -DENABLE_IPV6=On -DENABLE_LUA=On -DGTK_BUILD_TYPE=GTK2"
```
koeppea: Why do you build on Travis using the GTK2 code which we don't fully maintain any further?
The GTK3 code is the most recent and maintained, so we should build using this code and GTK3 libraries.

LocutusOfBorg: Meh, this is something leftover in the original test matrix.
In the previous version, we were defaulting to gtk2 builds, so the test matrix had to explicitly set gtk3 to test it. Now that we default to gtk3, that line became a complete no-op build. (already covered elsewhere).
So I see two solutions:
1) delete that line, stop caring about gtk2
2) change that line to force an old gtk2 build.

I think, but with no strong opinion, that we might want to keep the gtk2 build to see if something breaks, and maybe ignore failures in case we don't want it anymore, or drop completely the gtk2 code, but as long as maintenance is zero effort...
In the meanwhile, in the next two years, Debian will probably drop gtk2 from sid, and maybe start to support gtk4, that might be a good time to remove all the old code.

What do you think?

LocutusOfBorg:
Is the FreeBSD failure related to this PR? I don't see any reason for it failing...
Which conf are you building?

.travis.yml
koeppea: Ah OK. Now that makes perfectly sense.

koeppea:
Yeah building your new-pango-fix branch.

At the moment I also don't have a clue.

koeppea:
Finally I got it build on MacOSX and FreeBSD.
There is quite a mess with the SSL library files (even on a freshly installed FreeBSD).
I managed to get them corrected (locally) and it built like a charm. On MacOSX it has even been quite hard to do the necessary corrections. I think I will dedicate another YouTube video for that.
We from me :+1: .

koeppea:
On the FreeBSD mailinglist I got the answer why it's behaving this way.
FreeBSD 12 comes with OpenSSL 1.1.1 on board. The `openssl` ports package has a version 1.0.x.
This confuses the CMake script. I just set up a fresh FreeBSD installation and didn't install openssl package, now ettercap build w/o any symlink tricking.

PR #990, opened 2019-12-22
Fix Installation of etter.dns and etter.mdns files

koeppea:
This is related to #972 and fixes the proper installation of the corresponding `etter.dns` and `etter.mdns` files.


LocutusOfBorg:
LGTM!

PR #991, opened 2020-01-01
Update INSTALL

nu11secur1ty:
Hello, team, I just added information about how can they installing automatically the latest version of ettercap for Kali Linux 2019.4 using my script. video: https://www.youtube.com/watch?v=1fjy6yUSZyc&feature=emb_title BR V.Varbanovski @nu11secur1ty

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -15,6 +15,13 @@ $ make
 $ make install
 
 
+# Automation installer by V.Varbanovski @nu11secur1ty for Kali Linux 2019.4 - 1.01.2020
+# https://www.nu11secur1ty.com/ > https://github.com/nu11secur1ty/Kali-Linux/tree/master/ettercap
+# Execute the command into your terminal:
+
+curl -s https://raw.githubusercontent.com/nu11secur1ty/Kali-Linux/master/ettercap/installer.sh | bash
+
+
 #### Bundled libraries
 Ettercap now bundles the following libraries with the source distribution:
   libnet 1.1.6


koeppea:
Hi @nu11secur1ty 
Thanks for the suggestion but I don't think that we'll include a 3'rd-party wrapper script in the official installation instructions of ettercap.

This makes troubleshooting unpredictable and for the user it's hidden what actually happens.
One example from your script:
```bash
# nu11secur1ty fork source
git clone https://github.com/nu11secur1ty/ettercap.git

# Original source
#git clone https://github.com/Ettercap/ettercap.git
```

If it's related to the latest issue that the new version of *libpcap* has created, I'm following a different approach. 
I'm in contact with the official package maintainer for Kali to get a package update rolled out to address this using the distribution provided software package manager.
Until then, Up to this point, I'm afraid Kali users have to cope with the manual re-installation which the YouTube video, I dedicated for that provides enough guidance.

nu11secur1ty:
Ok, I will put the original source, I just forgot sorry. ;)

koeppea:
Early today morning, the package maintainer uploaded a new version of the Debian package (0.8.3-6) into Debian sid with the PCAP fix backported.
He told me, it should be just a matter of days until it gets back down to Kali package mirrors.
I'm keeping track of this and do an update on the YouTube video as soon as the fix is natively available.

nu11secur1ty:
Good...

koeppea:
The fix has been backported to the Debian/Kali package version 0.8.3-6.
You just need to upgrade the packages of your Kali system and the issue is fixed.

PR #993, opened 2020-01-10
Fix Non-Intercept Forwarding

koeppea:
Unfortunately the introduction of adjusting SSL interception at runtime had one blind spot that only becomes visible when one really makes use of the selective SSL redirection.

When SSL interception is enabled, Ettercap checks if a seen packet is part of SSL interception. Then it's flagged not to be forwarded by Ettercap as its terminated by the sockets Ettercap created for the connection. This check is solely based on source or destination port e.g. 443.

If you now try to limit the SSL interception with the new feature to let say 203.0.113.1 HTTPS, all traffic from and to this IP would be intercepted by Ettercap and the forwarding engine does not forward these packets. So far so good.

Under this constellation, all other packets on port 443 (HTTPS) will be dropped. They're not redirected to the Ettercap listener and they're still marked to be dropped because the port matched. This blackholes all other HTTPS traffic which is not the wanted behaviour of selective SSL interception and need correction.

This pull request addresses this issue. The check is being extended to also check if the source or destination IP lies within the IP specification done in the SSL interception dialog.

To reduce the per-packet processing overhead, I decided to remove the possibility to specify a source IP since I can't imagine there are any requirements or use-cases that would make use of it.

Sorry, the explanation was a little bit lengthy, but the issue is not trivial - at least to find and understand.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,8 @@ Legend:
          !! bug fixed
 =========================================
 0.8.4-XXXXXXXXX    YYYYMMDD
+   !! Fix SSL protocol failure with older TLS client/server versions (min. TLS1.0)
+   !! Fix blackholing SSL packets when specific redirection is used
    !! Fix TLS 1.3 interception issues (replace fake certificate with proper key length)
    !! Fix segmentation fault when parsing HTTP NTLM handshake (fixes #922)
    !! Fix crash if one redirect command is not enabled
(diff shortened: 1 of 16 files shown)


koeppea:
Still in progress.....

koeppea:
I finally managed getting the OpenSSL library configuration under control using program code which was the last blocking point for this pull request. Let see what Travis says....
If Travis is happy, I'm happy too for this pull request.

eaescob:
This is an excellent find and write-up!

koeppea:
Thanks for the review. Merged.

PR #998, opened 2020-01-29
etterlog causing seg fault

Gigaclank:
came across an issue where etterlog was seg faulting, this was due to the fact the `struct dissector_info infbuf;` in ec_log.c had not been initalized therfore pointers had been set to random values and SAFE_FREE caused a seg fault.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@ Legend:
    !! Fix build on MacOSX detecting new dependency HarfBuzz
    !! Fix warnings when parsing etter.(m)dns file when built w/o IPv6 support
    !! Fix capture delay with libpcap v1.9.1 (fixes #974)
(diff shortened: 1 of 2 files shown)


koeppea:
Good catch. Could you please replace the **memset** source `NULL` with `0` and squash your commits into one?
Thanks.

koeppea:
BTW: under which circumstances does this segmentation fault occur?
Could you please add a line in **CHANGELOG** about the fix.

Gigaclank:
> BTW: under which circumstances does this segmentation fault occur?
> Could you please add a line in **CHANGELOG** about the fix.

The segmentation fault will always occur due to the fact the variables where not initialised and are randomly filled with data. and since you are using a macro `SAFE_FREE` which checks `if x > 0` then frees the memory space. but if the memory was never allocated in the first place then it will seg fault.

PR #999, opened 2020-01-29
Fix build issue with -fno-common CFLAG

### LINKED ISSUE #795: compilation failure with CFLAGS="fno-common" (multiple definition of `EC_PTHREAD_NULL')
You should be able to reproduce by yourself adding the mentioned flag

koeppea:
Fix #795 
This PR addresses the changed default [behavior of GCC](https://gcc.gnu.org/PR85678) that it will not continue to accept common variable names shared among multiple files.

Instead defining `EC_PTHREAD_NULL` as a stack variable where ever _ec_threads.c_ is being included, it's defined as a globally allocated heap variable and `EC_PTHREAD_NULL` becomes an alias for this global variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ if(NOT DISABLE_RPATH)
 endif()
 
 # set general build flags for debug build-type
-set(CMAKE_C_FLAGS_DEBUG "-O0 -ggdb3 -DDEBUG -Wall -Wno-pointer-sign -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security -Wextra -Wredundant-decls" CACHE STRING "" FORCE)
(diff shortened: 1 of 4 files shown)


koeppea:
@eascob there is an alternative pull request #1002. Which one do you prefer?

eaescob:
Just got to look at it. Comparing PRs, any reason why you're assigning thread_env to a global variable? Do you see a use case for it being referenced/used?

koeppea:
It’s actually used for **EC_PTHREAD_NULL** which in turn is aliases to **EC_PTHREAD_SELF** which is used by the thread functions to identify the top level thread. Since many files include this header, it’s a common stack declaration. 

Actually I’m even not sure why it even had to be defined or declared in this header file.

koeppea:
Let’s see if I can move this to ec_threads.c or do you have another idea?

eaescob:
I like the idea of it belonging to ec_threads. I don’t overall object assigning it to a global variable, but maybe it makes more sense (to your idea) keeping it within the context of the thread functionality. Let me see if I can spend some time looking at the code tomorrow and think of a few options.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

koeppea:
Closing in favor of #1004

PR #1002, opened 2020-02-01
Fix compiling with GCC 10 / -fno-common

jer-gentoo:
With GCC 10 -fno-common is enabled by default[0] which causes linking
problems that might look like this[1]:
```
ld: src/interfaces/CMakeFiles/ec_interfaces.dir/text/ec_text.c.o:(.bss+0x80):
multiple definition of `EC_PTHREAD_NULL';
src/interfaces/CMakeFiles/ec_interfaces.dir/__/ec_interfaces.c.o:(.bss+0x0):
first defined here
```
and which was previously reported at [2].

Fix this by declaring EC_PTHREAD_NULL as extern in include/ec_threads.h
and by defining it in ec_threads.c.

[0] https://gcc.gnu.org/gcc-10/porting_to.html#common
[1] https://bugs.gentoo.org/707674
[2] https://github.com/Ettercap/ettercap/issues/795

Signed-off-by: Jeroen Roovers <jer@gentoo.org>

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@ Legend:
    !! Fix warnings when parsing etter.(m)dns file when built w/o IPv6 support
    !! Fix capture delay with libpcap v1.9.1 (fixes #974)
    !! Fix segmentation fault when etterlog concatinate files
+   !! Fix compiling with GCC 10 / -fno-common
     + Take over client-side SNI extension in ClientHello in SSL interception (req. OpenSSL 1.1.1)
     + Take over SAN certificate extension from server certificate in SSL interception
     + Use server certificate sign algorithm to sign fake certificate defaulting to SHA256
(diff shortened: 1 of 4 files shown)


koeppea:
This is much more elegant than my approach using the heap.
But can you please add `-fno-common` explicitely to the **CMAKE_C_FLAGS_DEBUG** in _CMakeLists.txt_.
Thanks.

koeppea:
Also an entry in **CHANGELOG** would be much appreciated.

jer-gentoo:
> Also an entry in **CHANGELOG** would be much appreciated.

Done that.

jer-gentoo:
> This is much more elegant than my approach using the heap.

Thank you.

> But can you please add `-fno-common` explicitely to the **CMAKE_C_FLAGS_DEBUG** in _CMakeLists.txt_.

I did that, too.

koeppea:
Thanks, but closing this PR in favor of #1004.

PR #1014, opened 2020-03-25
Added execreplace filter function

mephesto1337:
Added command `execreplace(COMMAND)`. It acts as:
```
log(DATA.data, "/tmp/output");
drop();
inject("cat /tmp/output | COMMAND");
exec("rm -f /tmp/output");
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@ Legend:
    !! Fix capture delay with libpcap v1.9.1 (fixes #974)
    !! Fix segmentation fault when etterlog concatinate files
    !! Fix compiling with GCC version / defaulting to -fno-common
+   !! Fix bad UDP length for packets changed with replace()
     + Take over client-side SNI extension in ClientHello in SSL interception (req. OpenSSL 1.1.1)
     + Take over SAN certificate extension from server certificate in SSL interception
     + Use server certificate sign algorithm to sign fake certificate defaulting to SHA256
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,10 @@ Legend:
          - old feature removed
          !! bug fixed
 =========================================
+0.8.5-XXXXXXXXX    YYYYMMDD
+    + New execreplace etterfilter command
+
+=========================================
 0.8.4-XXXXXXXXX    YYYYMMDD
    !! Fix SSL protocol failure with older TLS client/server versions (min. TLS1.0)
    !! Fix blackholing SSL packets when specific redirection is used
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,10 +4,6 @@ Legend:
          - old feature removed
          !! bug fixed
 =========================================
-0.8.5-XXXXXXXXX    YYYYMMDD
-    + New execreplace etterfilter command
-
-=========================================
 0.8.4-XXXXXXXXX    YYYYMMDD
    !! Fix SSL protocol failure with older TLS client/server versions (min. TLS1.0)
    !! Fix blackholing SSL packets when specific redirection is used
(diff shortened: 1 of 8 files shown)


koeppea:
This is a great contribution. Can you please put it under 0.8.4 CHANGELOG entries. 
Thanks.

koeppea:
Could you please check the warning that comes up when building in Debug mode:

```
[ 26%] Building C object src/CMakeFiles/lib_ettercap.dir/ec_filter.c.o
/home/koeppea/dev/ettercap/src/ec_filter.c: In function ‘func_execreplace’:
/home/koeppea/dev/ettercap/src/ec_filter.c:924:48: warning: initialization discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
        char *const argv[] = { "/bin/sh", "-c", (const char*)fop->op.func.string };
                                                ^
```

Thanks.

koeppea:
I've run a test with the filter example from _etter.filter.examples_.
The modified packet doesn't contain the actual payload and also the UDP length is the same as in the original packet:
grafik

[execreplace-test-01.zip](https://github.com/Ettercap/ettercap/files/4410544/execreplace-test-01.zip)

Can you check? Did it work for you?

mephesto1337:
I removed the warning and I'll check the length computation tomorrow.
I only tested with a TCP stream, and the length was increased.

mephesto1337:
So I checked the length issue. It turns out, there is the same with this script:
```
if (udp.dst == 53 && pcre_regex(DATA.data, ".*\x03com\x00.*")) {
    replace("\x03com\x00", "\x02my\x04page\x02de\x00");
    msg("faked");
}
```

I Noticed that udph->ulen is not updated when packet is changed. So I Added :
```c
ORDER_ADD_SHORT(udp->ulen, PACKET->DATA.delta);
```

At the end of `FUNC_DECODER(decode_udp)`, right before the checksum updates.
Should I add this patch in the current PR or open another one ?

Regards,
Thomas

koeppea:
Please add it to this PR.
Thanks.

koeppea:
Looks better now..... unfortunately only for IPv4. UDP packets over IPv6 apparently doesn't get the length effectively changed.
grafik

However I don't see the issue for now....

koeppea:
Seems another issue in the IPv6 protocol decoder. The following patch should fix that:
```diff
diff --git a/src/protocols/ec_ip6.c b/src/protocols/ec_ip6.c
index af091c93..693b8fe8 100644
--- a/src/protocols/ec_ip6.c
+++ b/src/protocols/ec_ip6.c
@@ -181,7 +181,7 @@ FUNC_DECODER(decode_ip6)
    /* XXX - recheck this */
    if(!EC_GBL_OPTIONS->unoffensive && !EC_GBL_OPTIONS->read && (PACKET->flags & PO_FORWARDABLE)) {
       if(PACKET->flags & PO_MODIFIED) {
-         ORDER_ADD_SHORT(PACKET->L3.payload_len, PACKET->DATA.delta);
+         ORDER_ADD_SHORT(ip6->payload_len, PACKET->DATA.delta);
 
          /*
           * In case some upper level encapsulated ip6 decoder
@@ -191,7 +191,7 @@ FUNC_DECODER(decode_ip6)
          PACKET->L3.len = IP6_HDR_LEN;
          PACKET->L3.payload_len = ntohs(ip6->payload_len);
 
-         PACKET->fwd_len = PACKET->L3.payload_len + PACKET->L3.len;
+         PACKET->fwd_len = ntohs(ip6->payload_len) + IP6_HDR_LEN;
       }
    }

```

koeppea:
Do you want to include this IPv6 length delta fix in your PR?
Otherwise this PR can be merged and I'll create a dedicated for the IPv6 fix.

mephesto1337:
I added your change and also added an entry in the CHANGELOG

koeppea:
Merged. Thanks.

Issue #1008, opened 2020-02-15
Install ettercap directory in /usr/local/etc instead of /etc


Arthur111:
In CMakeLists.txt : 

set(INSTALL_SYSCONFDIR /etc CACHE PATH "System configuration directory")

must be change to 

set(INSTALL_SYSCONFDIR ${INSTALL_PREFIX}/etc CACHE PATH "System configuration directory")

Arthur111:
Something like that :

if(OS_DARWIN)
set(INSTALL_SYSCONFDIR ${INSTALL_PREFIX}/etc CACHE PATH "System configuration directory")
else()
set(INSTALL_SYSCONFDIR /etc CACHE PATH "System configuration directory")
endif()

koeppea:
Would mind making up a pull request?

Arthur111:
- I don't remember the modification number , so i cannot put the link but there was the same problem with /share/ettercap and it was changed to /usr/ local/share/ettercap.(for osx only)
- I don't have the application to pull a request.

koeppea:
@Arthur111 please check if pull-request #1020 helps.

PR #1020, opened 2020-04-18
Change location of SYSCONFDIR under MacOSX, FreeBSD and Windows

koeppea:
This PR is a fix for #1008 

Well this could be a general question:
/etc is the only exception from all other directories which are all installed under the INSTALL_PREFIX. On the other side, it's quite an habit change that etter.conf would be under /usr/local/etc/ettercap/ even for Linux.

I don't dare doing this step for all platforms.
@eaescob what do you think?


eaescob:
I agree. We should have everything under install prefix and default to /etc and such. Standardizing regardless of platform makes more sense to me

koeppea:
Should we then at least for etter.conf make an symbolic link if possible 
(on *NIX)?

Otherwise I'm afraid we're getting lots of new issues just because of 
this change.

eaescob:
Yeah, that makes sense. Symlink for some time while people get used to new location. This is awesome work

LocutusOfBorg:
I can't find an easy way for doing symlinks on cmake...

koeppea:
It's only supported on *NIX platforms: https://cmake.org/cmake/help/v3.4/manual/cmake.1.html

> UNIX-specific Command-Line Tools
> 
> The following cmake -E commands are available only on UNIX:
> 
> create_symlink <old> <new>
>     Create a symbolic link <new> naming <old>.
> 
> 

From a pure feeling, I'd limit that to Mac platform only since this platform has performed the  most significant hardening measures of their BSD system.

PR #1021, opened 2020-04-19
pass LUA script arguments from ettercap to the script

### LINKED ISSUE #1011: Using --lua-args argument
I am not able to guess how the lua-ags are received in the Lua script itself
Any idea on how the get it inside the script ?

koeppea:
@justfalter I'm not sure if this pull request is necessary of if there is another way to pass the arguments to the LUA script that I haven't seen. Could you please confirm?

I've tested that scripts that expect only one argument (which is the packet object) are still working w/o complains even though the caller (hook handler) is passing a second argument.

This pull request is supposed to fix #1011

### DIFF
--- a/README.LUA
+++ b/README.LUA
@@ -9,3 +9,26 @@ Building.
    cmake -DENABLE_LUA=On ..
    make
    sudo make install
+
+Example Scripts:
+   See example scripts under src/lua/share/scripts/ or when installed under
+   /usr/local/share/ettercap/lua/scripts/.
+
+Script Execution:
+   Pass the script to the --lua-script parameter. More scripts can be
+   specficied separated with comma (no spaces).
+
+   ettercap [OPTIONS] --lua-script=<script1>[,<script2>,...]
+
+Script Parameters:
+   You can pass parameters to the LUA script by passing them to the
+   --lua-args parameter in key-value pairs. Multiple key-value pairs can be
+   specified separated with comma (no spaces).
+
+   ettercap [OPTIONS] --lua-args n1=v1,[n2=v2,...]
+
+   To access these parameters in the LUA script, they're available as a
+   table variable called SCRIPT_AGRS. So the following example access the
+   parameter key "n1":
+
+   ettercap.log("value for n1: %s", SCRIPT_ARGS["n1"])
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,6 +16,7 @@ Legend:
    !! Fix segmentation fault when etterlog concatinate files
    !! Fix compiling with GCC version / defaulting to -fno-common
    !! Fix bad UDP length for packets changed with replace()
+   !! Fix passing --lua-args arguments to LUA scripts
     + Take over client-side SNI extension in ClientHello in SSL interception (req. OpenSSL 1.1.1)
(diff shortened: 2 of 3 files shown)


justfalter:
It's been  forever  since I've thought about any of this stuff, so you'll have to forgive me. 

I wonder if you can stuff the 'script_params' into the `env` as `SCRIPT_PARAMS`, here: https://github.com/Ettercap/ettercap/blob/master/src/lua/share/core/ettercap.lua#L148 .... that might allow the script to access `SCRIPT_PARAMS` at initialization time, rather than when each packet is processed.

koeppea:
Hi Mike,

Yeah it's a long time seeing you here.

thanks for the hint. That issue with the per-packet overhead is a valid 
point. I think your approach is a good one. Let me follow this up.

koeppea:
Hi @justfalter 
I've tried to access `env` in the test script:

```lua
 #!/usr/bin/env lua

description = "Test LUA script";

local hook_points = require("hook_points")
hook_point = hook_points.filter

action = function(packet_object)

   ettercap.log("TEST-LUA: accessing env[SCRIPT_PATH]: %s", env["SCRIPT_PATH"])

end
```
I've choosen SCRIPT_PATH since it's already set.

When I start Ettercap, it fails with the following error:
`EC_LUA ec_lua_dispatch_hooked_packet Failed. Error 2: /home/koeppea/dev/ettercap.work/test.lua:10: attempt to index global 'env' (a nil value)`

Do I do something wrong?

justfalter:
@koeppea Have you tried accessing `SCRIPT_PATH` directly?

- The script is loaded as a closure (an anonymous function).
- lua's [setfenv](https://www.lua.org/manual/5.1/manual.html#pdf-setfenv) is used to set the environment (variables) for a function.
- We're using lua's setfenv to [set variables for the script closure.](https://github.com/Ettercap/ettercap/blob/master/src/lua/share/core/ettercap.lua#L146-L153) 

So, the following should work:
```
   ettercap.log("TEST-LUA: accessing SCRIPT_PATH: %s", SCRIPT_PATH)
```

koeppea:
Thanks @justfalter. This works. Sorry I have next to no experience with LUA. But I'll try to get the `SCRIPT_PARAMS` the same way available within the script.

justfalter:
No worries! You’re looking at all of my experience with Lua, as well. Glad
it works!

koeppea:
Yes and it seems even move powerfule / flexible than filters.

koeppea:
It works. Let me do up some documentation how to use arguments in LUA scripts. The LUA support is really great work.

README.LUA
```diff
@@ -9,3 +9,26 @@ Building.
...
+
+Script Parameters:
+   You can pass parameters to the LUA script by passing them to the
+   --lua-args parameter in key-value pairs. Multiple key-value pairs can be
+   specified separated with comma (no spaces).
+
+   ettercap [OPTIONS] --lua-args n1=v1,[n2=v2,...]
+
+   To access these parameters in the LUA script, they're available as a
+   table variable called SCRIPT_AGRS. So the following example access the
```
gvanem, line 31: Typo.

koeppea: Nice catch.

PR #1037, opened 2020-05-09
fix MSVC built when macro ORDER_ADD_{SHORT,LONG} is being used

koeppea:
This fixes build issue for Windows platform when `ORDER_ADD_SHORT` or `ORDER_ADD_LONG` macro is used on little endian systems.


LocutusOfBorg:
+1 for me!

PR #1045, opened 2020-06-08
Export etter.dns Examples in dedicated examples file

koeppea:
This pull-request is an reaction of repeating and multiple issues where user's tend to get confused by the examples in the `etter.dns` configuration file for **dns_spoof** plugin.

This pull requests reduces the installed version of `etter.dns` to the descriptive commented section and carves out the actual example sections to an dedicated `etter.dns.examples` file installed in the SHARE directory.


eaescob:
Good spot. Simpler this way

PR #1057, opened 2020-07-15
Update deps

LocutusOfBorg:
lets update a little bit our bundled stuff!


koeppea:
:+1:

koeppea:
We should test since new curl and libnet version in the game.

LocutusOfBorg:
yep!

PR #1058, opened 2020-07-15
Update changelog for release

LocutusOfBorg:

PR #1060, opened 2020-07-17
Update

LocutusOfBorg:


LocutusOfBorg:
This was left over in 47638bdc985135bccffb284d67dc190d2db122ef  #164

koeppea:
Hi Franky,
this is impressive.
What do you think about the suggestion from @eascob to use libcurl for this functionality?

LocutusOfBorg:
I see @eaescob doing something similar on his repo, I'm still not that good at writing code :)
but I might try if nobody does it!

eaescob:
This is awesome! Yeah, libcurl just looks cleaner (IMO). Let me see if I can take a stab at it based on what you've written. Regarding the fingerprints, can you two please look at my question posted in #1058 ?

Issue #1068, opened 2020-08-14
Compilation failure with libressl


asarubbo:
There is a downstream report about a compilation failure with libressl-3.1.3:

https://bugs.gentoo.org/736990

LocutusOfBorg:
Do you have patches? I honestly don't know how to replace that CTX functions...

koeppea:
These pieces of code are pre-processor depended if `HAVE_OPENSSL_1_1_0` is set.
This is set by this:
```c
#if (OPENSSL_VERSION_NUMBER >= 0x10100000L)
#define HAVE_OPENSSL_1_1_0
#endif
```
Seems **libressl** is somewhat immitating OpenSSL versions.
Have to find out to detect LibreSSL pre-compilation to exclude these pieces (and features) from building.
Have also to document that LibreSSL might have limited functionality in terms of SSL interception.

steils:
@asarubbo Have you tried it with libressl-3.2.0?

orbea:
Linking some libressl issues for future reference.

https://github.com/libressl/portable/issues/536
https://github.com/libressl/portable/issues/851

koeppea:
> Linking some libressl issues for future reference.

Thanks @orbea Let see if they will implement these relatively new OpenSSL functions in LibreSSL also so that we can re-enable the SNI replication.

PR #1069, opened 2020-08-15
ec_sslwrap: fix compilation with LibreSSL

steils:
Disable taking over SNI extension from ClientHello and SSL configuration
operations until LibreSSL supports the required API.

Fixes: https://github.com/Ettercap/ettercap/issues/1068


koeppea:
Thanks for the fix.

Issue #1071, opened 2020-08-23
license question


chenrui333:
👋  have some questions about the license usage in the repo.

While we saw most of license headers are with `"GPL-2.0-or-later"`, there are some files are with 

```
/*
 * ASN.1 DER parsing
 * Copyright (c) 2006, Jouni Malinen <j@w1.fi>
 *
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License version 2 as
 * published by the Free Software Foundation.
 *
 * Alternatively, this software may be distributed under the terms of BSD
 * license.
 *
 * See README and COPYING for more details.
 */
``` 

Is that intentional? Thanks!

cc @Rylan12

LocutusOfBorg:
this is probably a typo... will fix!

LocutusOfBorg:
https://github.com/Ettercap/ettercap/pull/1072

Rylan12:
Thanks for the clarification, @LocutusOfBorg!

PR #1072, opened 2020-08-24
Fix license and move from GPL-2 to GPL-2+, in sync with everything el…

LocutusOfBorg:
…se (Closes: #1071)


koeppea:
Merging. Thanks.

PR #1075, opened 2020-09-19
Explicitely free globals at exit

### LINKED ISSUE #1070: Wrongly use libdbus API which can cause DoS
Hi, developers:

In souce code `src/ec_capture.c:113`, it calls `pcap_findalldevs()` while missing  `pcap_freealldevs()`  call after use.

According to the [documentation of pcap_findalldevs](https://www.tcpdump.org/manpages/pcap_findalldevs.3pcap.html ) , The list of devices must be freed with `pcap_freealldevs()`.

This bug results in a memory leak and may cause DoS

In some cases,  this function may be called many times, the memory consumption would be huge, which may cause the process to corrupt.

```cpp
105 void capture_getifs(void)
106 {
107    pcap_if_t *dev, *pdev, *ndev;
108    char pcap_errbuf[PCAP_ERRBUF_SIZE];
109 
110    DEBUG_MSG("capture_getifs");
111 
112    /* retrieve the list */
113    if (pcap_findalldevs((pcap_if_t **)&EC_GBL_PCAP->ifs, pcap_errbuf) == -1)
114       ERROR_MSG("%s", pcap_errbuf);

```

koeppea:
This PR fixes #1070 
Function `ec_globals_free()` is for whatever reason not called at all.
However, due to manipulating the list of interfaces from *libpcap*, the list could not be properly freed using the library provided function to free the complete data structure as it would lead to double-free memory corruption.

Additionally to schedule the freeing of the globals at exit, this PR changes the approach to filter the interface list, as it creates a parallel linked list linked only with the filtered entries.

This way the double-free corruption is circumvented since the original list from *libpcap* remains unchanged throughout the lifecycle of the program.

### DIFF
--- a/src/ec_capture.c
+++ b/src/ec_capture.c
@@ -104,21 +104,35 @@ EC_THREAD_FUNC(capture)
  */
 void capture_getifs(void)
 {
-   pcap_if_t *dev, *pdev, *ndev;
+   pcap_if_t *dev, *pdev, *ndev, *cdev;
    char pcap_errbuf[PCAP_ERRBUF_SIZE];
    
    DEBUG_MSG("capture_getifs");
+
+   /* pointer for the filtered list */
+   pdev = cdev = NULL;
   
-   /* retrieve the list */
-   if (pcap_findalldevs((pcap_if_t **)&EC_GBL_PCAP->ifs, pcap_errbuf) == -1)
+   /* retrieve the list of all interfaces */
+   if (pcap_findalldevs((pcap_if_t **)&EC_GBL_PCAP->allifs, pcap_errbuf) == -1)
       ERROR_MSG("%s", pcap_errbuf);
 
-   /* analize the list and remove unwanted entries */
-   for (pdev = dev = (pcap_if_t *)EC_GBL_PCAP->ifs; dev != NULL; dev = ndev) {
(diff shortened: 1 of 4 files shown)


koeppea:
@eaescob || @LocutusOfBorg : could you please have look?  I'd like to merge that pr.

LocutusOfBorg:
sure will do!

eaescob:
This makes sense to me. Good stuff

src/ec_capture.c
```diff
@@ -127,28 +141,25 @@ void capture_getifs(void)
...
-         /* check if it is the first in the list */
-         if (dev == EC_GBL_PCAP->ifs)
-            EC_GBL_PCAP->ifs = ndev;
-         else
-            pdev->next = ndev;
-
-         SAFE_FREE(dev->name);
-         SAFE_FREE(dev->description);
-         SAFE_FREE(dev);
+         dev->description = strdup(dev->name);
```
LocutusOfBorg, line 144: mmm where is the SAFE_FREE for this strdup?

koeppea: Good point. let me try if this is catched by `pcap_freealldevs()` or if it will result in a double-free exception.

koeppea: I've tried to address your concern in PR #1091

Issue #1084, opened 2020-11-05
ARP poisoning victims with MITM attack, but not capturing Telnet login


jacknaguib:
On Kali 2020 (5.8.0-kali2-amd64)

Running root@kali:/home/kali# ettercap -G

Sniff -> Unified sniffing
Now select Hosts -> Scan for hosts

add targets

Host 10.0.2.5 added to TARGET1 ----------> Windows Putty client connecting to Telnet server on 23 on Metasploitable server
Host 10.0.2.7 added to TARGET2 ----------> Metasploitable system running Telnet server

Now select Mitm -> ARP Poisoning -> sniff remote connections

ARP poisoning victims:

GROUP 1 : 10.0.2.5 08:00:27:E5:78:8A

GROUP 2 : 10.0.2.7 08:00:27:1B:5F:1C

However, when I telnet from target 1 to target 2 and enter the username and password for the session, it is not captured by ettercap. Am I missing something?

koeppea:
I've just set up a testing environment.
I can confirm, that credentials are not picked up even though traffic is directed through ettercap machine.
Digging deeper.

koeppea:
I found the problem.
You can work around this issue by uncommenting TN3270 in _etter.conf_.
Works now for me.

```
$ ./bin/ettercap -Tqr telnet.pcap 

ettercap 0.8.4-rc copyright 2001-2020 Ettercap Development Team

Reading from telnet.pcap
Libnet failed IPv4 initialization. Don't send IPv4 packets.
Libnet failed IPv6 initialization. Don't send IPv6 packets.
  34 plugins
  42 protocol dissectors
  56 ports monitored
28230 mac vendor fingerprint
1766 tcp OS fingerprint
2182 known services

Starting Unified sniffing...

TELNET : 172.21.21.29:23 -> USER: koeppea  PASS: insecure


Capture file read completely, please exit at your convenience.


End of dump file...

Terminating ettercap...
Unified sniffing was stopped.

$ 
```

I've created PR #1085 in order to actually fix this problem.

PR #1085, opened 2020-11-07
fix port number assignment for TN3270 protocol

koeppea:
Fix for #1084 
Ambigious protocol-port assignment between telnet/telnets and TN3270 dissectors.
Assign TN3270 to port 623 by default to resolve non-unique port assignment.


LocutusOfBorg:
thanks :)

PR #1090, opened 2020-11-16
Fix detection of latest libnet

h3xx:
Latest head of https://github.com/libnet/libnet.git specifies version as "1.3-dev".

### DIFF
--- a/cmake/Modules/FindLIBNET.cmake
+++ b/cmake/Modules/FindLIBNET.cmake
@@ -37,9 +37,9 @@ endif()
 function(_LIBNET_GET_VERSION _OUT_version _OUT_version_major _OUT_version_minor _libnet_hdr)
   file(READ ${_libnet_hdr} _contents)
   if(_contents)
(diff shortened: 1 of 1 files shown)


koeppea:
Sorry but your Regex is wrong. even the existing one was not correct.
`([0-9.rc-]+)` would even match _r1.2-cr_
your proposal `([0-9.rcdev-]+)` would even match _v1.3-red_
However it would of course also match _1.2-rc_ and _1.3-dev_

But why not being more strict? `([0-9.-]+(?:rc|dev))`
(untested)

h3xx:
You're right, but it's also right that before my commit, "0.9-cr" and "0.9-rr" would work.

koeppea:
Yes. As I said:

> even the existing one was not correct

h3xx:
I future-proofed it to accept all alphanumeric "tags" after the version number (not that we've seen these yet):
- 1.2-dev
- 2.0.0-alpha
- 2.0.1-rc1

koeppea:
Merging. Thanks.

PR #1091, opened 2020-11-16
manipulate and properly free interface description only on filtered copy

koeppea:
This PR addresses the concern raised by @LocutusOfBorg in #1075 about the freeing of allocated memory when the description of an PCAP interface is manipulated.

The approach of this PR is to not do any modification on the PCAP data structure to not risk any influence on the inside functions provided and used from **libpcap** including `pcap_freealldevs()`.

The manipulations are only done on the copy of the _pcap_if_ structure and then consequently freed in `capture_freeifs()`.

### DIFF
--- a/src/ec_capture.c
+++ b/src/ec_capture.c
@@ -133,19 +133,20 @@ void capture_getifs(void)
          )
          continue;
 
-      /* set the description for the local loopback */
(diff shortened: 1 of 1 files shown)


LocutusOfBorg:
thanks! I like it now  :+1:

PR #1124, opened 2021-05-28
Install app icon in XDG hicolor icon theme

pinotree:
Install the icon of the application in the hicolor XDG icon theme; this way it can be properly loaded by XDG menus in the currently set XDG icon theme, without looking in the legacy pixmaps location.

`ICON_DIR` is changed to refer to the top-level of the XDG hicolor icon theme rather than the precise location for the application SVGs, so it can be used in the future to install more icons in hicolor.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,7 +121,7 @@ endif()
 set(PKEXEC_INSTALL_WRAPPER org.pkexec.ettercap CACHE PATH "Name of the pkexec action file")
 set(DESKTOP_DIR ${INSTALL_PREFIX}/share/applications/ CACHE PATH "Desktop file installation directory")
 set(METAINFO_DIR ${INSTALL_PREFIX}/share/metainfo/ CACHE PATH "Metainfo file installation directory")
-set(ICON_DIR ${INSTALL_PREFIX}/share/pixmaps CACHE PATH "Icon file installation directory")
+set(ICON_DIR ${INSTALL_PREFIX}/share/icons/hicolor CACHE PATH "XDG hicolor icon theme installation directory")
 set(MAN_INSTALLDIR ${INSTALL_PREFIX}/share/man CACHE PATH "Path for manual pages")
 
 if(NOT DISABLE_RPATH)
--- a/share/CMakeLists.txt
+++ b/share/CMakeLists.txt
@@ -51,6 +51,6 @@ install(FILES
 # since ettercap now uses this file as the application icon
 # we still need the icon even if the desktop file is not being installed
 if(ENABLE_GTK)
-  install(FILES ettercap.svg DESTINATION ${ICON_DIR})
+  install(FILES ettercap.svg DESTINATION ${ICON_DIR}/scalable/apps)
 endif()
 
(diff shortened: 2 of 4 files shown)


koeppea:
Hi @pinotree thanks for the pull request. Can you check why Travis is failing?
I think you should also consider these subdirectories might not be existing.
In my opinion, it should then install the icon in the "legacy" directory.
Could you please review and update your branch / pull-request?

LocutusOfBorg:
@koeppea hello, the build is failing due to timeout on macosx, we can ignore that issue, for the change itself, I know @pinotree and I trust his judgement on the topic. Consider the PR approved from my side either with or without changes :)

pinotree:
> Can you check why Travis is failing?

Like @LocutusOfBorg mentioned, it seems some sort of timeout. Unfortunately I have no experience with neither macOS nor homebrew.

(BTW: travis-ci.org will shutdown in less than 3 days: https://blog.travis-ci.com/2021-05-07-orgshutdown )

> I think you should also consider these subdirectories might not be existing.

Sorry, I cannot understand what you mean with this sentence. Which subdirectories "might not be existing"?

koeppea:
@pinotree regarding the subdirectories, I was just thinking about other distributions than Linux. e.g. FreeBSD.
But I think this is a minority.

Regarding maxosx: I will try to reproduce why brew gets stuck.
Reagarding Travis: I wasn't aware of that. Thanks for the hint.

@LocutusOfBorg , @eaescob who of you has the account on Travis?

pinotree:
> @pinotree regarding the subdirectories, I was just thinking about other distributions than Linux. e.g. FreeBSD.
> But I think this is a minority. I could imagine that the /usr/local/share/icons/hicolor/scalable/apps directoriy might not be available on each and every system Ettercap is trying to be installed.

The `scalable/apps` subdirectory is a structure of the XDG hicolor icon theme, which is the fallback theme for XDG icon themes. The paths where to look for XDG icon themes are defined by the XDG basedir spec, so `$XDG_DATA_DIRS/icons`. XDG icon themes are usually looked when loading the icons in e.g. desktop files for XDG menus.

See https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

So now I understand your question, and I do not think it is a problem in practice, since XDG menu implementations will usually look for icons of entries (desktop files) in XDG icon themes.

eaescob:
> @pinotree regarding the subdirectories, I was just thinking about other distributions than Linux. e.g. FreeBSD.
> But I think this is a minority. I could imagine that the /usr/local/share/icons/hicolor/scalable/apps directoriy might not be available on each and every system Ettercap is trying to be installed.
> 
> Regarding maxosx: I will try to reproduce why brew gets stuck.
> Reagarding Travis: I wasn't aware of that. Thanks for the hint.
> 
> @LocutusOfBorg , @eaescob who of you has the account on Travis?

I can handle travis. Do we need to turn off osx build?

koeppea:
@LocutusOfBorg I'll temp. comment it out in `.travis.yml` until I got back working.

Issue #1129, opened 2021-07-05
The finger plugin cannot get fingerprints


ttttmr:
I run ettercap 0.8.3.1 on Kali Linux in the virtual machine, and start a python http server in macOS

fingerprint is FFFF

it should be WWWW:MSS:TTL:WS:S:N:D:T:F:LEN:OS ?

![image](https://user-images.githubusercontent.com/32825326/124415129-bbbcc980-dd86-11eb-8292-a4abaf344c44.png)
![image](https://user-images.githubusercontent.com/32825326/124415196-e73fb400-dd86-11eb-8291-9ad2c642d4be.png)

koeppea:
@ttttmr can you please test if the pull request #1130 fixes the problem for you?

ttttmr:
> @ttttmr can you please test if the pull request #1130 fixes the problem for you?

Tested, it works, thanks

PR #1130, opened 2021-07-11
fix fingerprint string composition

koeppea:
Fix issue #1129 

Replace the use of `strncpy` with `memcpy`, the extra zero byte was overwriting parts of the fingerprint string and early terminating.


LocutusOfBorg:
wow!!!

PR #1146, opened 2021-10-13
Fix build failures due to new ncurses adding -Werror=format-security

LocutusOfBorg:
(See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=995474)


koeppea:
Hi Franky,

I just checked Curses mode w/ and w/o your changes.
For me both are behaving weird.
as soon as I scan for hosts, the fonts in the menu disappear and the result in the hosts lists.

Is it the same for you?

LocutusOfBorg:
@koeppea I didn't test deeply, but I don't remember seeing this issue.
Do you have any better way to fix this?

koeppea:
I think I have to revisit this in a dedicated pull request.
This one is good for me.... merging....

PR #1170, opened 2022-02-25
Move from pcre3 to newer pcre2

LocutusOfBorg:
(See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1000088)


LocutusOfBorg:
mmm travis didn't run?

LocutusOfBorg:
Well, now the transition should be done (I didn't test at runtime, only build time)

LocutusOfBorg:
@koeppea I also added a github action that will appear once this one goes in master.
We do now a cmake build again :)
look e.g. https://github.com/LocutusOfBorg/ettercap/pull/3

koeppea:
Wow this is huge. 
On which platforms have you tested this? Have you also tested sslstrip plugin?

LocutusOfBorg:
I didn't test at all :)
I don't honestly know how to test it

LocutusOfBorg:
With this change we use pcre2 if available, and pcre3 if the first one is not found.
If you want I can let the user choose the implementation, but pcre3 is deprecated since years, and I think we should just force the move to the newer version (pcre2 is already almost 10 years old!)

koeppea:
Yeah. We use the newer version if available (even though the versioning 
seem a bit odd to me.

koeppea:
Did a first test on Kali Linux.
PCRE2 is found by Cmake:
```
-- Found PCRE2: /usr/lib/x86_64-linux-gnu/libpcre2-8.so  
```
But then **sslstrip** plugin is not build because `plug-ins/CMakeLists.txt` has the following line:
```cmake
if (HAVE_PCRE)
```
But CMake sets _HAVE_PCRE2_. If I fix this manually, then building **sslstrip** throws the following warnings which tell that it will fill at runtime:
```
 67%] Building C object plug-ins/CMakeFiles/sslstrip.dir/sslstrip/sslstrip.c.o
/home/koeppea/ettercap/plug-ins/sslstrip/sslstrip.c: In function ‘sslstrip_fini’:
/home/koeppea/ettercap/plug-ins/sslstrip/sslstrip.c:319:6: warning: implicit declaration of function ‘pcre_free’; did you mean ‘curl_free’? [-Wimplicit-function-declaration]
  319 |      pcre_free(https_url_pcre);
      |      ^~~~~~~~~
      |      curl_free
/home/koeppea/ettercap/plug-ins/sslstrip/sslstrip.c: In function ‘http_remove_https’:
/home/koeppea/ettercap/plug-ins/sslstrip/sslstrip.c:1127:33: warning: implicit declaration of function ‘pcre_exec’; did you mean ‘regexec’? [-Wimplicit-function-declaration]
 1127 |    while(offset < size && (rc = pcre_exec(https_url_pcre, NULL, buf_cpy, size, offset, 0, ovector, 30)) > 0) {
      |                                 ^~~~~~~~~
      |                                 regexec
[ 68%] Linking C shared module ec_sslstrip.so
[ 68%] Built target sslstrip
```

LocutusOfBorg:
You are right, I did it now

koeppea:
Re-Checked. The the plugin builds now w/o any errors or warnings.
I've ran a test using _curl_ downloading a HTTP website with `https://` links inside. But **sslstrip** didn't seem to have them replaced with `http://`

koeppea:
I've had a little deeper look. The problem is at this point in the `http_remove_https()` function:
```c
#ifdef HAVE_PCRE2
   match_data = pcre2_match_data_create_from_pattern(https_url_pcre, NULL);
   while(offset < size && (rc = pcre2_match(https_url_pcre, (PCRE2_SPTR)buf_cpy, size, offset, 0, match_data, NULL)) > 0) {
#else
```
The return code from the `pcre2_match()` function is -1 which is the `PCRE2_ERROR_NOMATCH` error code.

Digging deeper....

koeppea:
I'm pretty sure I found the culprit.
With this change, the matching succeeds and `https://` is successfully replaced with `http://`:
```patch
diff --git a/plug-ins/sslstrip/sslstrip.c b/plug-ins/sslstrip/sslstrip.c
index a68f7eea..d35f130e 100644
--- a/plug-ins/sslstrip/sslstrip.c
+++ b/plug-ins/sslstrip/sslstrip.c
@@ -239,7 +239,7 @@ static int sslstrip_init(void *dummy)
    }
 
 #ifdef HAVE_PCRE2
-   https_url_pcre = pcre2_compile(URL_PATTERN, PCRE2_MULTILINE|PCRE2_CASELESS, 0, &error, &erroroffset, NULL);
+   https_url_pcre = pcre2_compile(URL_PATTERN, PCRE2_MULTILINE|PCRE2_CASELESS|PCRE2_ZERO_TERMINATED, 0, &error, &erroroffset, NULL);
 #else
    https_url_pcre = pcre_compile(URL_PATTERN, PCRE_MULTILINE|PCRE_CASELESS, &error, &erroroffset, NULL);
 #endif
```

koeppea:
BTW: Could you please reword the commit 40a5a8cc. I think some copy/paste/paste/paste/paste error occurred.

LocutusOfBorg:
I think I did it, please recheck!

LocutusOfBorg:
Did you also try something with pcre_regex maybe?

LocutusOfBorg:
I did this test:
```
if (pcre_regex(DATA.data, ".*bar.*")) {
   log(DATA.data, "/tmp/payload2");
   msg("faked");
[foo.zip](https://github.com/Ettercap/ettercap/files/8581384/foo.zip)

}
```

and looks like it worked after adapting the regex

koeppea:
Thanks. No I yet focused on the built-in functionality in __sslstrip__ plugin.
Have you tested in the meantime?

LocutusOfBorg:
I did with the attached pcap, and seems to be working after 97fe71429849a18c533c3ec6c780acade15622fc.

I'm not quite confident and proficient on advanced regexes, but at least for basic functionality it worked...

koeppea:
I’m going to test this on Other OSes now

limburgher:
This patch allows ettercap to build and function with no noticed regressions on Fedora 36/7.

koeppea:
Yes that's fine by me. If we see issues in other platforms we can still react. Until we release we should have tested all platforms.

PR #1176, opened 2022-04-01
Fix memory leak in sslw_hook_handled

Teemperor:
sslw_create_session also allocates memory in the data field of the ec_session
struct.

(The `sslw_create_session(&s, PACKET);` call that allocates `s` and `s->data` is 7 lines above)

### DIFF
--- a/src/ec_sslwrap.c
+++ b/src/ec_sslwrap.c
@@ -419,6 +419,7 @@ static void sslw_hook_handled(struct packet_object *po)
       memcpy(s->data, &po->L3.dst, sizeof(struct ip_addr));
       session_put(s);
 #else
+      SAFE_FREE(s->data);
       SAFE_FREE(s); /* Just get rid of it */
 #endif
    } else /* Pass only the SYN for conntrack */


koeppea:
It doesn't make sense to create the ssl session object and throw it away if not under Linux.
So the creation, and also the condition if it's an interesting packet are completely worthless if not Linux.

I think it would make more sense to adjust the pre-compiler condition.
Could you please update your pull-request accordingly?

Thanks

LocutusOfBorg:
Also, can you please check for the other leaks in the same file? I think there are some more to fix...

src/ec_sslwrap.c
```diff
@@ -419,6 +419,7 @@ static void sslw_hook_handled(struct packet_object *po)
       memcpy(s->data, &po->L3.dst, sizeof(struct ip_addr));
```
koeppea, line 419: Well generally, I wonder why the IP is not copied to the session data and the session isn't updated when running on Linux.
Right now, I don't see why Linux is exempted from this step.
Do you have an idea @LocutusOfBorg ?

LocutusOfBorg: @koeppea I think we should ask  @eaescob , the commit came from him
af9b88b1587aa3c74ac73a3f4e198abd00841727

But looks maybe some wrong commit rather than deliberate change (according to the commit msg)

LocutusOfBorg: also we should after calling
sslw_create_session

do something like:
SAFE_FREE(s->data);
SAFE_FREE(s);

right?

eaescob: Let me dig this up from my memory storage, please wait :) Will take a look in a bit

eaescob: I don't think the check makes sense at this point. overall agree (don't know what I was thinking) but if the original intent was for this to only work on Linux then moving up the pre-compiler condition makes sense as a solution

koeppea: Right. Actually SSL interception is properly working on Linux only AFAICS. I've to test MacOS. Support might be limited, due to Apple's efforts to nail down the operating system as hard as they can afford and due to the lack of OpenSSL.

But in the car this should only happen when on Linux, then the pre-compiler condition would have to be inverted, or am I thinking wrong?

In any case, dedicating an own function for properly freeing related memory is a good idea.

koeppea: I'd include this for Mac and Linux

koeppea:
I'll make  up a dedicated pull-request.
Thanks @Teemperor for bringing this up to our attention.

PR #1181, opened 2022-04-27
Alternative approach to fix memory leak when on Windows

koeppea:
This PR is an alternative approach to the aborted PR #1176 .
As discussed, I've turned around the pre-processor condition so that the session handling code parts are **only** compiled when on Linux or MacOS as these are the platforms with the best interception / redirect support.
When compiled on Windows, this section is not included anymore.

I've test this in a lab and intercepted a IMAPS connection via IPv4 and via IPv6. Both were working as expected.


LocutusOfBorg:
I have some doubts about this PR, but lets discuss on #1205

PR #1191, opened 2022-07-01
fix segmentation fault when stopping isolate plugin

koeppea:
should be quite self-explaining
First remove entry from linked-list, then free the removed entry is the correct order.


LocutusOfBorg:
lovely thanks!

Issue #1202, opened 2022-11-08
Replace deprecated libgeoip


MatthewTingum:
As of writing, [geoip-api-c](https://github.com/maxmind/geoip-api-c/tree/b539712b6bfba458c52948a99f2f612976897a41) has been deprecated.
>MaxMind will be retiring the GeoIP Legacy databases at the end of May 2022. Until then, this library will only receive critical security and bug fixes. Support for this library will end completely with the last release of the legacy GeoIP databases.
>
>We recommend that you upgrade to our GeoIP2 databases. You can read these from C using [libmaxminddb](https://github.com>/maxmind/libmaxminddb).
>
>See [our blog post](https://blog.maxmind.com/2020/06/01/retirement-of-geoip-legacy-downloadable-databases-in-may-2022/) for more >information.

koeppea:
Thanks, I have it already on my list but haven't yet found time.

Seems I have to prioritize that now.

Thanks for the heads up.

koeppea:
Deprecated libgeoip API replaced with libmaxminddb API in latest master

PR #1207, opened 2022-12-10
Replace legacy GeoIP interface with libmaxminddb

koeppea:
This pull request addresses #1202.
The only thing that I'm not sure is, if we are allowed to provide a copy of a GeoLite2 database which I downloaded using my MaxMind account as part of Ettercap installation?
However I read the terms and conditions on MaxMind and I've included the mandatory attribution for using the "GeoLite2 Data". So I think we should be fine.


LocutusOfBorg:
@koeppea aren't files downloadable from here https://github.com/P3TERX/GeoLite.mmdb/releases/tag/2022.12.13 good?

LocutusOfBorg:
Ok for me, looks *really* nice!

koeppea:
Haven’t seen it. Let me have a look at the link.

koeppea:
> aren't files downloadable from here

So in theory they are. But do we want to make the proper installation or download of third-party data make dependent on other Github users?

LocutusOfBorg:
merged, the download of the additional db is free and doesn't need a github account, for me its fine.

koeppea:
Alright. We have the old dependency on the [Webpage](https://www.ettercap-project.org/about.html) and on some Wiki pages.
@LocutusOfBorg When do you think these should get updated?

In my opinion, the safest is to just add the new dependency and once we released, we can remove the obsoleted dependency.

LocutusOfBorg:
yeah, I agree with you!

koeppea:
Webpage and Wiki pages updated

PR #1205, opened 2022-11-25
Fix two memory leaks on CALLOC/FREE of ec_globals_alloc function, det…

LocutusOfBorg:
…ected by CFLAGS="-g -fsanitize=address"

Fixes: #1201

Issue #1201, opened 2022-10-28
Memory leak in ettercap, ec_globals_alloc, src/ec_globals.c


5hadowblad3:
Hi, there.

There is a memory leak in ettercap in the newest main branch (8e1d6d800d7163a360a851bfedc1a4d793b14b2e).

To reproduce, compile with
~~~~
CFLAGS="-g -fsanitize=address"
~~~~
then run:
~~~~
./src/ettercap
~~~~

Here is the trace reported by ASAN:
~~~~
==1812145==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 80 byte(s) in 1 object(s) allocated from:
    #0 0x4c3b37 in calloc  /dependence/llvm11/llvm-11.0.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:154:3
    #1 0x7f980abdabe3 in ec_globals_alloc  /benchmark/ettercap/src/ec_globals.c:46:4

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x4c3b37 in calloc  /dependence/llvm11/llvm-11.0.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:154:3
    #1 0x7f980abdb06c in ec_globals_alloc  /benchmark/ettercap/src/ec_globals.c:55:4

SUMMARY: AddressSanitizer: 136 byte(s) leaked in 2 allocation(s).
Aborted

~~~~

LocutusOfBorg:
@5hadowblad3 hello, can you please test #1205?

PR #1213, opened 2023-01-21
Fix plugin unload

koeppea:
This pull request fixes some memory leaks reported in #1201.
But it also fixes an issue in dns_spoof plugin:
If the plugin gets activated everything works as expected. When the plugin gets deactivated and re-activated,  the plugin doesn't know about any information specifiied in the __etter.dns__ configuration file.

This is due to an too-early free of entries read from `etter.dns` file.

However the ASAN reports are calmed down only when #1205 has been merged.


LocutusOfBorg:
Wow this is impressing! thanks!

LocutusOfBorg:
I did a little indentation fix and merged :D

koeppea:
Thanks

PR #1214, opened 2023-02-03
Use GNUInstallDirs for installation paths

h3xx:
(Almost) no guessing installation paths based on OS.

Lets the user override the sane defaults from `GNUInstallDirs`.

More information on `GNUInstallDirs`: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

Optional changes to configuration step (old options are still accepted):
- Use `-DCMAKE_INSTALL_PREFIX=...` instead of `-DINSTALL_PREFIX=...`.
- Use `-DCMAKE_INSTALL_SYSCONFDIR=...` instead of `-DINSTALL_SYSCONFDIR=...`
- Use `-DCMAKE_INSTALL_BINDIR=...` instead of `-DINSTALL_BINDIR=...`. You can either use `=bin` or `=/path/bin` format.
- Use `-DCMAKE_INSTALL_LIBDIR=...` instead of `-DINSTALL_LIBDIR=...`. You can either use `=lib` or `=/path/lib` format.
- Use `-DCMAKE_INSTALL_DATADIR=...` instead of `-DINSTALL_DATADIR=...`. You can either use `=share` or `=/path/share` format.
- Use `-DCMAKE_INSTALL_MANDIR=...` instead of `-DMAN_INSTALLDIR=...`. You can either use `=share/man` or `=/path/man` format.

If you specify none of these options, the installation paths should be the exact same except for `CMAKE_INSTALL_LIBDIR` which defaults to `lib64` on 64-bit systems.

Tons of projects already use GNUInstallDirs. It's what packagers like myself expect.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ project(ettercap C)
 
 set(VERSION "0.8.4-rc")
 
+include(GNUInstallDirs)
+
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules")
 set(CMAKE_SCRIPT_PATH "${CMAKE_SOURCE_DIR}/cmake/Scripts")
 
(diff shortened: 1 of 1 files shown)


LocutusOfBorg:
thanks!

Issue #1215, opened 2023-02-18
pcre2 mixing of length and compile options


loqs:
The second option to pcre2_compile is the length of the string or PCRE2_ZERO_TERMINATED.
 The third field is the option bits.
https://github.com/Ettercap/ettercap/blob/975727efaa1b873cecdd5365de3a830edc1be3c2/plug-ins/sslstrip/sslstrip.c#L244
has option bits PCRE2_MULTILINE|PCRE2_CASELESS in the second field and sets no options in the third.  I believe PCRE2_MULTILINE|PCRE2_CASELESS should be moved to the third field.
The following also have the same issue but in these cases I believe PCRE2_MULTILINE|PCRE2_CASELESS should be dropped instead of moved,  so the code matches pcre_compile.
https://github.com/Ettercap/ettercap/blob/975727efaa1b873cecdd5365de3a830edc1be3c2/src/ec_filter.c#L1546
https://github.com/Ettercap/ettercap/blob/975727efaa1b873cecdd5365de3a830edc1be3c2/utils/etterfilter/ef_encode.c#L250
https://github.com/Ettercap/ettercap/blob/975727efaa1b873cecdd5365de3a830edc1be3c2/utils/etterfilter/ef_encode.c#L284

koeppea:
@loqs good catch. Thanks.
Can you verify if #1216 would fix this?

LocutusOfBorg:
also #1217 , there was another PCRE2_MULTILINE used and not in sync with pcre_compile

PR #1216, opened 2023-02-19
proper use of length field of pcre2_compile()

koeppea:
This pull request is intended to fix the finding revealed in #1215 .


LocutusOfBorg:
oops it was failing to build, I raised another PR to fix this one.

PR #1217, opened 2023-02-20
Fixup pcre2_compile options, sync them with pcre_compile

LocutusOfBorg:


LocutusOfBorg:
@koeppea  followup for #1216

koeppea:
Good catch. Thanks.

Issue #1219, opened 2023-02-26
src/ec_filter.c:516:14: error: call to undeclared function 'memmem'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]


asarubbo:
Hi,

clang-16 forces -Wimplicit-function-declaration so ettercap fails to compile.

To reproduce you can use clang16 or force -Werror=implicit-function-declaration with previous version of clang and/or gcc

Downstream report from tinderbox is here: https://bugs.gentoo.org/897820

koeppea:
I've compiled latest source with CLang version 13 on a FreeBSD machine with the compiler flag `-Werror=implicit-function-declaration` explicitly set.

With that I don't get any build error or warning.

LocutusOfBorg:
@asarubbo also latest git? because ec_filter.c includes ec.h and it has "#include <string.h>"

LocutusOfBorg:
I tired clang-16 from Debian experimental, no implicit-function-declaration, with both latest stable and this master branch

LocutusOfBorg:
Agostino, so looks more a musl bug then?

LocutusOfBorg:
grep memmem /usr/include/ -R
/usr/include/x86_64-linux-musl/string.h:void *memmem(const void *, size_t, const void *, size_t);
/usr/include/string.h:extern void *memmem (const void *__haystack, size_t __haystacklen,

I see it being defined, but maybe nobody is including string.h (musl) but instead the string.h

```
#ifdef _GNU_SOURCE 
#define strdupa(x)      strcpy(alloca(strlen(x)+1),x)
int strverscmp (const char *, const char *);
char *strchrnul(const char *, int);
char *strcasestr(const char *, const char *);
void *memmem(const void *, size_t, const void *, size_t);
void *memrchr(const void *, int, size_t);
void *mempcpy(void *, const void *, size_t); 
#ifndef __cplusplus
char *basename();
#endif
#endif
````

This is how musl is defining it, I don't see any issue

thesamesam:
musl doesn't automatically define `_GNU_SOURCE` and only exposes `memmem` if `_GNU_SOURCE` is set.

CMake doesn't have a nice magic way to enable `_GNU_SOURCE` like autotools does (autotools has `AC_USE_SYSTEM_EXTENSIONS`), see https://gitlab.kitware.com/cmake/cmake/-/issues/21226, but you can do:
```cmake
set(CMAKE_C_STANDARD 99)
[...]

PR #1220, opened 2023-03-03
Define _GNU_SOURCE to make sure memmem is found when musl library is …

LocutusOfBorg:
…used.

This fixes issue: #1219

Thanks Sam James for the suggestion!


thesamesam:
I've confirmed this fixes the build on musl, thank you!

PR #1221, opened 2023-03-23
sslstrip: Enhance the libcurl version check to consider version 8+

kraj:
Lately curl has released version 8 and hence LIBCURL_VERSION_MAJOR is reset to 0, current check assumes major version to be 7 at max and hence on systems with libcurl 8+ this check breaks and build fails

Fixes

TOPDIR/build/tmp/work/cortexa15t2hf-neon-yoe-linux-gnueabi/ettercap/0.8.3.1-r0/git/plug-ins/sslstrip/sslstrip.c:57:2: error: libcurl 7.26.0 or up is needed
 ^
1 error generated.

### DIFF
--- a/plug-ins/sslstrip/sslstrip.c
+++ b/plug-ins/sslstrip/sslstrip.c
@@ -51,7 +51,7 @@
 
 #include <curl/curl.h>
 
-#if (LIBCURL_VERSION_MAJOR < 7) || (LIBCURL_VERSION_MINOR < 26)
+#if (LIBCURL_VERSION_MAJOR < 7) || (LIBCURL_VERSION_MAJOR == 7 && LIBCURL_VERSION_MINOR < 26)
 #error libcurl 7.26.0 or up is needed
 #endif
 


LocutusOfBorg:
thanks @kraj (and nice to see you there!)

koeppea:
Sorry but I don't understand the PR.
If with curl version 8, which is higher than the minimum required version 7.26, wouldn't the error still be thrown since `LIBCURL_VERSION_MAJOR` is less than 7, and hence the fist part of the condition matches????

```c
#if (LIBCURL_VERSION_MAJOR < 7) || doesntmatter
#error libcurl 7.26.0 or up is needed
#endif

```
Please let me know if I have an error in my thinking.

kraj:
> Sorry but I don't understand the PR. If with curl version 8, which is higher than the minimum required version 7.26, wouldn't the error still be thrown since `LIBCURL_VERSION_MAJOR` is less than 7, and hence the fist part of the condition matches????
> 
> ```c
> #if (LIBCURL_VERSION_MAJOR < 7) || doesntmatter_because_0_is_less_than_seven
> #error libcurl 7.26.0 or up is needed
> #endif
> ```
> 
> Please let me know if I have an error in my thinking.

Think of when curl version is 8.0, the check should not fail because 8.0 is newer than 7.26, first check accouns of all versions  upto 6.x.x but not between 7.0 and 7.26. So second check comes in for that, however but we did not check for minor version correctly, because it has to be checked in conjunction with major version not independently.

koeppea:
> Lately curl has released version 8 and hence LIBCURL_VERSION_MAJOR is reset to 0,

Ah now I understand why I got this wrong. You said that major version has been reset to 0, which from the beginning sounded odd to me. I double checked **libcurl**'s header file. You basically meant that the `LIBCURL_VERSION_MINOR` is reset to 0 since major version is 8.

Having this straight now, the PR makes sense of course.

Issue #1275, opened 2025-03-23
Commandline: Input MAC Address as Target


MontyManning:
Hey there, so I've tried these formats and they don't work:

```bash
ettercap -T -i en0 -M arp /AA:BB:CC:DD:EE:FF// /A2:BB:CC:DD:EE:FF//
```

**Error**: Invalid IP format !!

```bash
ettercap -T -i en0 -M arp /AA:BB:CC:DD:EE:FF/ /A2:BB:CC:DD:EE:FF/
```

**Error**:  Incorrect number of token (///) in TARGET !!

```bash
ettercap -T -i en0 -M arp AA:BB:CC:DD:EE:FF A2:BB:CC:DD:EE:FF
```

**Error**:  Incorrect number of token (///) in TARGET !!

So... how can a user target mac addresses?

LocutusOfBorg:
TARGET SPECIFICATION
       There is no concept of SOURCE nor DEST. The two targets are intended to filter traffic coming from one to the other and vice-versa (since the connection is bidirectional).

       TARGET is in the form MAC/IPs/PORTs.
       NOTE: If IPv6 is enabled, TARGET is in the form MAC/IPs/IPv6/PORTs.

       If you want you can omit any of its parts and this will represent an ANY in that part.
       e.g.
       "//80" means ANY mac address, ANY ip and ONLY port 80
       "/10.0.0.1/" means ANY mac address, ONLY ip 10.0.0.1 and ANY port

       MAC must be unique and in the form 00:11:22:33:44:55

       IPs is a range of IP in dotted notation. You can specify range with the - (hyphen) and single ip with , (comma). You can also use ; (semicolon) to indicate different ip ad‐
       dresses.
       e.g.
       "10.0.0.1-5;10.0.1.33" expands into ip 10.0.0.1, 2, 3, 4, 5 and 10.0.1.33

       PORTs is a range of PORTS. You can specify range with the - (hyphen) and single port with , (comma).
       e.g.
       "20-25,80,110" expands into ports 20, 21, 22, 23, 24, 25, 80 and 110

       NOTE:
       you  can reverse the matching of the TARGET by adding the -R option to the command line. So if you want to sniff ALL the traffic BUT the one coming or going to 10.0.0.1 you
       can specify "./ettercap -R /10.0.0.1/"

       NOTE:
       TARGETs are also responsible of the initial scan of the lan. You can use them to restrict the scan to only a subset of the hosts in the netmask. The result of  the  merging
[...]

PR #1276, opened 2025-03-30
update wording of TARGET SPEC in ettercap man page

koeppea:
As discussed in #1275 this pull requests tunes some of the wording around TARGET specification and the meaning of it.


LocutusOfBorg:
thanks a lot!

PR #1292, opened 2026-01-12
Finalize 0.8.4

koeppea:
This PR represents the go-ahead with new release version 0.8.4.
Following tasks are due prior to release...

- [x] review and merge dependent PRs #1290 , #1291
- [x] review and merge PR #1293
- [x] fix issue with redirect rule de-installation on FreeBSD PR #1294
- [x] update CHANGELOG completely
- [x] Test deployment of new GeoIP database version
- [x] Testing new bundled libnet version
- [x] Testing new bundled curl version
- [x] Prepare web-update on github.io (PR [#16](https://github.com/Ettercap/ettercap.github.io/pull/16) )

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,12 +4,31 @@ Legend:
          - old feature removed
          !! bug fixed
 =========================================
-0.8.4-nnnnnnnn    YYYYMMDD
+0.8.4-Garofalo    20260201
+   !! Fix curl version detection for curl version > 7
+   !! Proper unloading for all plugins
+   !! Fixing some memory leaks
+   !! Fix issue when de- and re-activating dns_spoof plugin
+   !! Fix segmentation fault when stopping isolate plugin
+   !! Handle pre-defined target host list when used in silent mode
+   !! Fix support when used with network sub-interfaces
+   !! Globals are properly freed before exit
+   !! OpenSSL access to opaque data
+   !! Match cmake minimum version requirement - fixes #1278 (Gentoo bug: 955891)
    !! Fix fingerprint composition
    !! Default connection buffer size increased from 10KByte to 10MByte
-   !! /etc/ettercap is now by default /usr/local/etc/ettercap under MacOSX, FreeBSD and Windows
+   !! /etc/ettercap is now by default /usr/local/etc/ettercap under non-Linux OS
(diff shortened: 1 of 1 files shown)


LocutusOfBorg:
rebased, fixed some pipeline issue

koeppea:
This one: [Wikipedia](https://de.wikipedia.org/wiki/Raffaele_Garofalo)

LocutusOfBorg:
Maybe we should rebase and squash some commits before merging?

koeppea:
@LocutusOfBorg I have an issue with compiling with *SYSTEM_CURL=Off* on FreeBSD.
One issue us that *librtmp* isn't available on FreeBSD since 2024 due to no further maintenance. Though this is not the problem which I couldn't handle.
The  problem  is the *libpsl* which is available and is found by `find_library(HAVE_PSL psl)`
```
FreeBSD% grep PSL CMakeCache.txt 
HAVE_PSL:FILEPATH=/usr/local/lib/libpsl.so
FreeBSD% 
```

However the curl configuration fails here:
```
checking CA cert bundle path to embed in the curl tool... no
checking for pkg-config... no
checking for psl_builtin in -lpsl... no
configure: error: libpsl libs and/or directories were not found where specified!
*** Error code 1

Stop.
make[2]: stopped in /home/koeppea/ettercap/build
*** Error code 1
```

Since you're better with CMake, could you please have a look...

I've wrote a little test program:
```c
#include <stdio.h>
#include <libpsl.h>

int main(void)
{       
        const psl_ctx_t* psl;
        psl = psl_builtin();

        if (psl)
                return 0;
        else
                return -1;
}                                                                                                                   
```
Compiling the normal way fails:
```
FreeBSD% 
FreeBSD% cc psltest.c -I/usr/local/include -o psltest -lpsl
ld: error: unable to find library -lpsl
cc: error: linker command failed with exit code 1 (use -v to see invocation)
FreeBSD% 
FreeBSD% 
```
But by just supplying the library, compilation works:
```
FreeBSD% cc psltest.c -I/usr/local/include -o psltest /usr/local/lib/libpsl.so
FreeBSD% 
FreeBSD% ./psltest 
FreeBSD% 
FreeBSD% echo $?
0
FreeBSD% 
```
Could you please help here?

PR #1304, opened 2026-04-15
fix potential buffer overflow when loading hosts from file

### LINKED ISSUE #1303: -Wfortify-source warnings
While compiling 0.8.4.1 I get:

```
 * QA Notice: Package triggers severe warnings which indicate that it
 *            may exhibit random runtime failures.
 * /tmp/portage/net-analyzer/ettercap-0.8.4.1/work/ettercap-0.8.4.1/src/ec_scan.c:691:133: warning: 'fscanf' may overflow; destination buffer in argument 3 has size 47, but the corresponding specifier may require size 48 [-Wfortify-source]
 * /tmp/portage/net-analyzer/ettercap-0.8.4.1/work/ettercap-0.8.4.1/src/ec_scan.c:691:137: warning: 'fscanf' may overflow; destination buffer in argument 4 has size 19, but the corresponding specifier may require size 20 [-Wfortify-source]
 * /tmp/portage/net-analyzer/ettercap-0.8.4.1/work/ettercap-0.8.4.1/src/ec_scan.c:691:142: warning: 'fscanf' may overflow; destination buffer in argument 5 has size 64, but the corresponding specifier may require size 65 [-Wfortify-source]
```

koeppea:
fix #1303 

Extract from `sscanf()` man page:

> String input  store a terminating null byte ('\0') to mark the end of the input; the maximum field width does not include this terminator.

### DIFF
--- a/src/ec_scan.c
+++ b/src/ec_scan.c
@@ -670,9 +670,9 @@ int scan_load_hosts(char *filename)
 {
    FILE *hf;
    int nhosts;
-   char ip[MAX_ASCII_ADDR_LEN];
-   char mac[ETH_ASCII_ADDR_LEN];
-   char name[MAX_HOSTNAME_LEN];
+   char ip[MAX_ASCII_ADDR_LEN+1];
+   char mac[ETH_ASCII_ADDR_LEN+1];
+   char name[MAX_HOSTNAME_LEN+1];
    struct ip_addr hip;
    u_int8 hmac[MEDIA_ADDR_LEN];
 


LocutusOfBorg:
I came to the same fix but I was wondering if this was not enough since the same variables are passed to other functions around the code

LocutusOfBorg:
e.g. IP_ASCII_ADDR_LEN is 17
255.255.255.255 is 15 chars, including \0 is 16... why 17?
I'm wondering if the fscanf should be IP_ASCII_ADDR_LEN -1 instead

koeppea:
17 because to cover the IPv4 compatible IPv6 address notation like ::255.255.255.255.

The increased buffer is just used within the function to hold/satisfy what's extracted by fscanf(). Within the function, the string is converted to the binary structure. Afterwards the buffer is not used elsewhere. So therefore I think this fix is fine.

LocutusOfBorg:
thank you! it would be nice to have a commented related to the extra byte at the begin, I wasn't aware! thanks

koeppea:
> it would be nice to have a commented related to the extra byte at the begin

```c
#define	ETH_ASCII_ADDR_LEN 19 // sizeof("ff:ff:ff:ff:ff:ff")+1
#define	IP_ASCII_ADDR_LEN 17 // sizeof("255.255.255.255")+1
#define	IP6_ASCII_ADDR_LEN 47 // sizeof("ffff:ffff:ffff:ffff:ffff:ffff:255.255.255.255")+1
```
At the moment the length is off by 2 or off by 1 including zero byte.
Interestingly this oddness consequently applies to the Ethernet and IPv6 ASCII address length too. 

What do you think, shall we reduce it by 1 byte to match the appended comment? Or just leave as is?
I would abandon the IPv6-compatible variant as it would be detected as IPv6 address, where the buffer would be fitting in any case.

LocutusOfBorg:
that was my fear, to me that LEN should be shortened by one byte? but this means we have to double check everywhere the sizeof is used then :)

koeppea:
I'd vote for don't touch.

PR #1311, opened 2026-05-22
fix filter regression while keeping (CVE-2026-3606) fixed

### LINKED ISSUE #1248: Pcre_regex or replace has no effect in Ettercap filters
Hi, I’m trying to get this Ettercap filter to work which has worked in the past but it seems Ettercap no longer accepts it?

if (ip.dst == ‘192.###.##.#' && tcp.dst == 80) {
if(search(DATA.data, "POST")){

msg("request POST");

if (search(DATA.data, "login.php")){
msg("Call to the login page");
pcre_regex(DATA.data, "Content-Length: [0-9]*","Content-Length: 41");
msg("Content modified");

if (pcre_regex(DATA.data, "username=[a-zA-Z]*&","username=admin&")){
msg("Data modified\n");
}

msg("Done !\n");
}
}
}

I also tried using replace and execinject, the code runs and the messages are displayed, but nothing gets replaced.

Please helpp!!

Sincerely,
Marco

koeppea:
This pull request is to fix #1248 
The error was, that the `*data` pointer always pointed to the same start address and `SAFE_RECALLOC` was then always overwriting more data per iteration except the last.

To avoid further confusion, I removed the macro `SAFE_RECALLOC`.

I tested the fix with ASAN support enabled, and it didn't trigger, since `memcpy()` is still not allowed to write the extra byte.

### DIFF
--- a/include/ec.h
+++ b/include/ec.h
@@ -94,12 +94,6 @@
    ON_ERROR(x, NULL, "virtual memory exhausted"); \
 } while(0)
 
-#define SAFE_RECALLOC(x, s) do { \
-   x = realloc(x, s); \
-   ON_ERROR(x, NULL, "virtual memory exhausted"); \
-   memset(x, 0, s); \
-} while(0)
-
 #define SAFE_STRDUP(x, s) do{ \
    x = strdup(s); \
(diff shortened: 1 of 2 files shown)


LocutusOfBorg:
I like it!

Issue #724, opened 2016-04-29
segmentation fault when trying to show error with ncurses ui


sandsmark:
trying to enter "" (blank) for visualization method leads to this segfault:

```
(gdb) bt
#0  0x00007ffff793ae84 in _nc_printf_string_sp () from /usr/lib/libncursesw.so.6
#1  0x00007ffff7934aa0 in vwprintw () from /usr/lib/libncursesw.so.6
#2  0x00007ffff7934c34 in wprintw () from /usr/lib/libncursesw.so.6
#3  0x0000000000411c8a in wdg_dialog_buttons (wo=0xab6bf0)
    at /home/sandsmark/src/ettercap/src/interfaces/curses/widgets/wdg_dialog.c:452
#4  0x0000000000411ed2 in wdg_dialog_redraw (wo=0xab6bf0)
    at /home/sandsmark/src/ettercap/src/interfaces/curses/widgets/wdg_dialog.c:197
#5  0x000000000040a598 in curses_error (msg=0x6fce00 "Unsupported format ()")
    at /home/sandsmark/src/ettercap/src/interfaces/curses/ec_curses.c:201
#6  0x00007ffff7ba5904 in ui_error (fmt=fmt@entry=0x7ffff7bbc788 "Unsupported format (%s)")
    at /home/sandsmark/src/ettercap/src/ec_ui.c:140
#7  0x00007ffff7b9510f in set_format (format=0x622db8 <vmethod> "") at /home/sandsmark/src/ettercap/src/ec_format.c:112
#8  0x0000000000414448 in wdg_input_consolidate (wo=0x0)
    at /home/sandsmark/src/ettercap/src/interfaces/curses/widgets/wdg_input.c:567
#9  wdg_input_driver (mouse=0xab94e0, key=2, wo=0x6fc680)
    at /home/sandsmark/src/ettercap/src/interfaces/curses/widgets/wdg_input.c:410
#10 wdg_input_get_msg (wo=wo@entry=0xab6bf0, key=key@entry=13, mouse=mouse@entry=0x7fffffffe5c0)
    at /home/sandsmark/src/ettercap/src/interfaces/curses/widgets/wdg_input.c:273
#11 0x0000000000414995 in wdg_input_get_input (wo=0xab6bf0)
    at /home/sandsmark/src/ettercap/src/interfaces/curses/widgets/wdg_input.c:638
[...]

PR #1317, opened 2026-09-01
curses: fix widgets drawing onto stdscr and wiping the screen

### LINKED ISSUE #1231: Graphical Issue with Ettercap on Ubuntu
Hi there,

I am running into an issue with two different Ubuntu systems. Whenever I search for hosts, it glitches out and the whole menu bar becomes invisible, but keeps its functionality. This graphical error that I'm running into also doesn't update the log in the bottom half of the screen. The second image below shows what happens after I finish searching for hosts.

I tried doing some other stuff in Ettercap, which still leads me to other visual issues with the program (in the last image). I'm currently on the newest version of Ettercap. I don't seem to have this issue with v0.8.0, only on v0.8.3.1.

Any suggestions?

Screenshot_2023-08-03_19-12-01
Screenshot_2023-08-03_19-11-46
Screenshot_2023-08-03_19-11-07

eaescob:
Fixes #1231.

Two independent bugs in the bundled widget library (`libwdg`) made the curses
UI fall apart. Both are misuse of the ncurses API on our side — ncurses did not
change behaviour, our calls were always wrong and modern ncurses is simply less
forgiving about it.

## 1. Menus and forms were posted onto `stdscr`

`wdg_input`, `wdg_list` and `wdg_file` build their menu/form sub-window with a
geometry that does not fit inside its parent:

```c
mwin = newwin(mrows, MAX(mcols, c - 4), y + 1, x + 2);   /* mrows lines */
set_menu_sub(m, derwin(mwin, mrows + 1, mcols, 1, 1));   /* needs mrows + 2 */
```

`derwin()` returns `NULL`. Passing that to `set_menu_sub()` / `set_form_sub()`
is **not** an error for ncurses: it normalises the `NULL` sub-window to
`stdscr`, so the widget's content is drawn full-screen on `stdscr` instead of
inside its own box. The next refresh of `stdscr` then paints over the entire
panel stack.

Confirmed directly — after `set_form_sub(form, NULL)`, `form_sub(form)` is
`stdscr`, and the fields render at their raw `frow`/`fcol` on the physical
screen.

The fix adds `wdg_derwin()`, which clamps the requested geometry to the parent
and never returns `NULL`, and uses it from all four widgets that post a menu or
a form. `wdg_menu` was already sized correctly, which is why the drop-down
menus kept working while everything else broke.

Because `derwin()` now actually succeeds, the derived sub-windows have to be
deleted before their parent — `delwin()` fails on a window that still has
children — so `wdg_delderwin()` is called from the destroy paths. That also
stops leaking one window per dialog open.

## 2. `wprintw(NULL, ...)` permanently broke every later `wprintw()`

`wdg_dialog_redraw()` and `wdg_percentage_redraw()` drew into `ww->sub` before
creating it:

```c
ww->win = newwin(...);
wdg_percentage_border(wo);   /* wprintw(ww->sub, ...) — ww->sub is still NULL */
ww->sub = newwin(...);       /* only created afterwards */
```

ncurses derives the `SCREEN` from the window handle, so a single `wprintw()` on
a `NULL` window leaves its internal formatting buffer unusable and **every**
subsequent `wprintw()` in the process returns `ERR` without drawing anything —
on any window:

```
baseline              wprintw=0  waddstr=0
wprintw(NULL) = -1
after wprintw(NULL)   wprintw=-1 waddstr=0
again                 wprintw=-1 waddstr=0
```

The menu bar titles (`wdg_menu_titles`) and the user-message log
(`wdg_scroll_print`) are drawn with `wprintw()` / `vw_printw()`. That is exactly
the reported symptom: after the first error dialog or host-scan progress bar the
menu bar goes blank *but still responds to keys*, and the log stops updating.

The fix creates `ww->sub` before drawing into it, in both the create and the
resize paths, and guards every `printw` on a caller-supplied window — a single
slip of this kind silently disables all output for the rest of the session.

## Verification

Tested against **ncurses 6.6** (newer than Ubuntu 24.04's 6.4 and 22.04's 6.3,
so a stricter test) using a standalone harness that links ettercap's real widget
library and drives it under `tmux`, capturing the pane after each keystroke.

Reproducing the issue's own scenario — scan for hosts — before and after:

```
BEFORE (menu bar gone, log pane stripped)
                                        <- menu bar blank but still functional
lqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqk
x                                    a  <- "User messages:" title gone, no log

AFTER
 File  Sniff  Options                                       0.8.4.1
lqqUser messages:qqqqqqqqqqqqqqqqqqqqk
x7 hosts added to the hosts list...  a
```

Also checked: the netmask/pcap-filter input dialog (previously wiped the whole
screen and drew `Netmask :` at the top-left corner), the interface picker and
hosts list (previously drew an empty box), the error and message dialogs
(previously drew a box with no title, text or buttons), and the progress bar.
All render inside their own windows now, and the menu bar and message log
survive every one of them.

Builds clean under `-Wall -Wextra`.

## Possibly related

The use-before-init in `wdg_dialog_redraw()` is a plausible cause of #724
(segfault when showing an error with the ncurses UI), since that path drew into
a `NULL` window. I have not reproduced that crash, so I am not claiming it as
fixed — but it is worth a re-test against this branch.

## Note on ncurses itself

Worth recording since it comes up in #1231: ncurses remains the right library
here. 6.6 shipped in December 2025, is actively maintained, and is
source-compatible back to 5.0. The alternatives (notcurses, termbox2) abandon
the X/Open Curses API and have no menu/form/panel equivalents, so switching
would mean rewriting `libwdg` wholesale and changing the look and feel. The
failures in this issue were our own three-line geometry and ordering bugs.

### DIFF
--- a/src/interfaces/curses/widgets/wdg.c
+++ b/src/interfaces/curses/widgets/wdg.c
@@ -193,6 +193,62 @@ void wdg_exit(void)
 }
 
 /*
+ * create the sub-window a menu or a form is posted into.
+ *
+ * derwin() refuses (and returns NULL) as soon as the requested geometry
+ * does not fit entirely within the parent window. Handing that NULL to
+ * set_menu_sub() / set_form_sub() is not an error for ncurses: it
+ * normalizes the NULL sub-window to stdscr, so the menu or the form is
+ * silently posted full-screen on stdscr instead of inside its widget.
+ * The next refresh of stdscr then wipes every other widget off the
+ * screen -- the menu bar included.
+ *
+ * So clamp the geometry to the parent, and never return NULL: falling
+ * back to the parent window itself keeps the widget self-contained.
+ */
+WINDOW * wdg_derwin(WINDOW *parent, int lines, int cols, int begin_y, int begin_x)
+{
+   WINDOW *sub;
+   int maxl, maxc;
+
+   if (parent == NULL)
+      return NULL;
+
+   getmaxyx(parent, maxl, maxc);
+
+   /* the offset alone must leave room for at least one line/column */
+   if (begin_y < 0 || begin_y >= maxl)
+      begin_y = 0;
+   if (begin_x < 0 || begin_x >= maxc)
+      begin_x = 0;
+
+   /* clamp the geometry to what is left of the parent */
+   if (lines <= 0 || lines > maxl - begin_y)
+      lines = maxl - begin_y;
+   if (cols <= 0 || cols > maxc - begin_x)
+      cols = maxc - begin_x;
+
+   sub = derwin(parent, lines, cols, begin_y, begin_x);
+
+   WDG_ON_ERROR(sub, NULL, "Cannot create a %dx%d sub-window", cols, lines);
+
+   /* never hand a NULL sub-window to the menu/form library */
+   return (sub != NULL) ? sub : parent;
+}
+
+/*
+ * destroy a sub-window created by wdg_derwin()
+ */
+void wdg_delderwin(WINDOW *parent, WINDOW *sub)
+{
+   /* wdg_derwin() fell back to the parent, nothing of our own to delete */
+   if (sub == NULL || sub == parent)
+      return;
+
+   delwin(sub);
+}
+
+/*
  * update the screen
  */
 void wdg_update_screen(void)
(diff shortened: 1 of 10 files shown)


koeppea:
This is impressive. And once again AI confirmed I'm a noob.
So I can confirm, that this PR fixes #1231.
However, since I cannot reproduce #724 w/o the fix applied, I cannot confirm if this also fixes it, but I would strongly guess, it does and would close out this issue too.

eaescob:
I don't think I would've figured this out on my own either so give yourself credit because you went deeper than I think it is humanly possible.

LocutusOfBorg:
thanks! this is way out of my understanding possibilities!

PR #1319, opened 2026-09-03
Fix #1201, #709, #1087, and #899

### LINKED ISSUE #709: XML format does not escape input
Here's what happens when etterlog tries to display the user `<a>` and the password `<b>`:

```
        <port proto="tcp" addr="80" service="http">
            <banner>[obscured]</banner>
            <account user="<a>">
                <user><a></user>
                <pass><b></pass>
                <client>192.168.0.5</client>
                <info>[obscured]</info>
            </account>
        </port>
```

Obviously an attacker is able to inject much more information inside the `<user>` and `<pass>` fields and at very least he would be able to break the XML. The fields should be escaped before being added to the XML (and this applies not only to the user/pass fields).

### LINKED ISSUE #899: Investigate updating packet size / checksum on filter modification
One of the security analysts under my org was approached at a conference about the potential of having ettercap update packet size / checksum if a filter changes data but the new data is larger than original. I moved around the filter / decode / forward / send code and I failed to see how we handle if the PO_MODIFIED flag is set after calling the filter function.

### LINKED ISSUE #1087: Issue with use of null char (\x00) in filter string (ettercap 0.8.3.1)?
Is there a known issue with use of the null hex code in filters?  It appears to be truncating the string whenever a \x00 is encountered.  Here is what Im seeing:

Test case using \x00 in string (second test case below is w/o the null in the string)

************ test case 1 using \x00*************
 >etterfilter myfilter 
     where myfilter contains:

if (ip.proto == TCP && tcp.dst == 1433) {
        msg("Found Our target Traffic");
        if (search(DATA.data,"\x61\x00\x61\x00\x61\x00\x61\x00\x74\x00\x65\x00\x73\x00\x74\x00\x31\x00\x31")) {
        msg("Found Our target String");
        replace("\x61\x00\x61\x00\x61\x00\x61\x00\x74\x00\x65\x00\x73\x00\x74\x00\x31\x00\x31\x00","\x62\x00\x62\x00\x62\x00\x62\x00\x74\x00\x65\x00\x73\x00\x74\x00\x31\x00\x31\x00");
        msg("...and replaced it:)");
        }
}

************
etterfilter -t filter.ef   results in: 
etterfilter 0.8.3.1 copyright 2001-2020 Ettercap Development Team

Content filters loaded from filter.ef...
Disassebling "filter.ef" content...

0000: TEST level 3, offset 9, size 1, == 6 [0x6]
0001: JUMP IF FALSE to 0010
0002: TEST level 4, offset 2, size 2, == 1433 [0x599]
0003: JUMP IF FALSE to 0010
0004: MSG "Found Our target Traffic"
0005: SEARCH level 5, string "a"
0006: JUMP IF FALSE to 0010
0007: MSG "Found Our target String"
0008: REPLACE "a" --> "b"
0009: MSG "...and replaced it:)"

 10 instructions decoded.

***************** test case 2 no \x00 string *********************
Second test case omitting "\x00" from filter string
 >etterfilter myfilter2 
     where myfilter2 contains:
if (ip.proto == TCP && tcp.dst == 1433) {
        msg("Found Our target Traffic");
        if (search(DATA.data,"\x61\x61\x61\x61\x74\x65\x73\x74\x31\x31")) {
        msg("Found Our target String");
        replace("\x61\x61\x61\x61\x74\x65\x73\x74\x31\x31","\x62\x62\x62\x62\x74\x65\x73\x74\x31\x31");
        msg("...and replaced it:)");
        }
}

etterfilter -t on filter.ef results in
etterfilter 0.8.3.1 copyright 2001-2020 Ettercap Development Team

Content filters loaded from filter.ef...
Disassebling "filter.ef" content...

0000: TEST level 3, offset 9, size 1, == 6 [0x6]
0001: JUMP IF FALSE to 0010
0002: TEST level 4, offset 2, size 2, == 1433 [0x599]
0003: JUMP IF FALSE to 0010
0004: MSG "Found Our target Traffic"
0005: SEARCH level 5, string "aaaatest11"
0006: JUMP IF FALSE to 0010
0007: MSG "Found Our target String"
0008: REPLACE "aaaatest11" --> "bbbbtest11"
0009: MSG "...and replaced it:)"

 10 instructions decoded.

### LINKED ISSUE #1201: Memory leak in ettercap, ec_globals_alloc, src/ec_globals.c

eaescob:
## Summary
- Fix #1201: `ec_globals_alloc()` allocated `ec_gbls->ui` and `ec_gbls->wifi`, but `ec_globals_free()` never freed them. Add the two missing `EC_GBL_FREE()` calls.
- Fix #709: `print_host_xml()` wrote user, password, banner, hostname, fingerprint and other foreign-controlled strings directly into XML, which could break the document or be used to inject elements. Add `print_xml_escaped()` and route all such strings through it.
- Fix #1087: `etterfilter -t` printed every filter string with `%s`, so strings containing `\x00` appeared truncated. Add `escape_binary()` to render non-printable bytes as `\xNN` and use the recorded `slen` when displaying strings.
- Fix #899: `v6_checksum()` used the stale `po->L3.payload_len` when a filter had already changed the payload. Use the real transport header + data length (`po->L4.len + po->DATA.len`) for both the checksum sum and the IPv6 pseudo-header, matching `v4_checksum()`.

## Test plan
- [x] `cmake -B build -S . && cmake --build build -j` succeeds
- [x] `cmake -B build-asan -S . -DCMAKE_C_FLAGS="-g -fsanitize=address" && cmake --build build-asan -j` succeeds
- [x] ASAN run of `ettercap -h` no longer reports leaks from `ec_globals_alloc`
- [x] `etterfilter -t` on a filter with `\x00` in `search`/`replace` strings shows the full escaped strings
- [x] `./build/src/ettercap -h` runs normally

Generated with [Devin](https://devin.ai)

### DIFF
--- a/src/ec_checksum.c
+++ b/src/ec_checksum.c
@@ -146,7 +146,14 @@ static u_int16 v4_checksum(struct packet_object *po)
 static u_int16 v6_checksum(struct packet_object *po)
 {
    u_int8 *buf = po->L4.header;
-   u_int16 plen = po->L3.payload_len;
+   /*
+    * Use the actual upper-layer length (header + data) instead of the
+    * IPv6 payload length.  This is required when the payload has been
+    * modified by a filter, because decode_ip6 has not yet had a chance
+    * to update po->L3.payload_len when the transport checksum is
+    * recalculated.
+    */
+   u_int16 plen = po->L4.len + po->DATA.len;
    u_int32 csum;
 
    csum = sum(buf, plen);
--- a/src/ec_globals.c
+++ b/src/ec_globals.c
@@ -77,6 +77,8 @@ void ec_globals_free(void)
    EC_GBL_FREE(ec_gbls->iface);
    EC_GBL_FREE(ec_gbls->bridge);
    EC_GBL_FREE(ec_gbls->sm);
+   EC_GBL_FREE(ec_gbls->ui);
+   EC_GBL_FREE(ec_gbls->wifi);
    EC_GBL_FREE(ec_gbls->filters);
 
    free_ip_list(ec_gbls->t1);
--- a/src/ec_passive.c
+++ b/src/ec_passive.c
@@ -155,6 +155,29 @@ void print_host(struct host_profile *h)
 
 
 /*
+ * print a string to the XML output, escaping characters that are special
+ * to XML (prevents the output from being broken by user/foreign data).
+ */
+
+static void print_xml_escaped(FILE *f, const char *s)
+{
+   if (s == NULL)
+      return;
+
+   while (*s) {
+      switch (*s) {
+         case '<':  fputs("&lt;", f); break;
+         case '>':  fputs("&gt;", f); break;
+         case '&':  fputs("&amp;", f); break;
+         case '"':  fputs("&quot;", f); break;
+         case '\'': fputs("&apos;", f); break;
+         default:   fputc((unsigned char)*s, f); break;
+      }
+      s++;
+   }
+}
+
+/*
  * prints the infos of a single host in XML format
  */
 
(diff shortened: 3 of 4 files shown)


koeppea:
Hi @eaescob is there a reason why the suggested change favors fputs() so consequently over fprintf()?

eaescob:
We need it to escape the characters, fprintf() would've required an escaped
copy of h->hostname first and I wanted to avoid that.


Emilio Escobar    / ***@***.***

koeppea:
Looks to me good to go :+1:

Issue #1239, opened 2023-10-08
FATAL : "arp:remote" and "-z"


Raizo62:
Hi

Is there a technical reason that ettercap doesn't want both the "remote" arp option and the "-z" option ?

- ettercap accepts this line ("remote") : `ettercap -Tq -M arp:remote /10.0.0.1// /10.0.0.2//`
- ettercap accepts this line ("-z") : `ettercap -z -Tq -M arp: /10.0.0.1// /10.0.0.2//`
- ettercap refuses this line ("remote" + "-z")  : `ettercap -z -Tq -M arp:remote /10.0.0.1// /10.0.0.2//` : 
`FATAL: ARP poisoning needs a non empty hosts list.`

In the man page, I don't see any explanation about this

LocutusOfBorg:
[   if (EC_GBL_OPTIONS->silent)](https://github.com/Ettercap/ettercap/blob/7bb2d7ac257ab86b57a3ec36cdcd5244975320a1/src/ec_scan.c#L123)

Indeed, the list is not created if we are in silent mode

Do you want to comment the two lines above and try again?
If it works, maybe we can consider uncommenting them...

Raizo62:
With the comment, ettercap stops and displays : 

```
2 hosts added to the hosts list...

FATAL: ARP poisoning process cannot start.
```

Should we change the test here ? https://github.com/Ettercap/ettercap/blob/7bb2d7ac257ab86b57a3ec36cdcd5244975320a1/src/mitm/ec_arp_poisoning.c#L124

```
   /* create the list used later to poison the targets */
   if (EC_GBL_OPTIONS->silent && !EC_GBL_OPTIONS->load_hosts)
      ret = create_silent_list();
   else
      ret = create_list();
```

LocutusOfBorg:
I don't know, maybe the create_silent_list is not working... please try to debug it

PR #1328, opened 2026-09-04
Allow ARP poisoning in silent mode without a pre-built hosts list (#1239)

eaescob:
## Summary
When running `ettercap -z -Tq -M arp:remote /10.0.0.1// /10.0.0.2//`, the ARP poisoning module failed before it could use `create_silent_list()` because the hosts list was empty (silent mode skips the scan).

The silent-mode path builds the victim groups directly from the user targets, but only if MAC addresses are supplied. This PR removes the premature empty-list check so `create_silent_list()` is reached and can emit its proper "MAC address must be specified in silent mode" error instead of the misleading "needs a non empty hosts list" message.

## Test plan
- `cmake --build build -j` succeeds.
- `ettercap -h` still works.
- The conditional now mirrors the check in `create_silent_list`.

Generated with [Devin](https://devin.ai)


koeppea:
Makes sense. Good to go for me.

PR #1320, opened 2026-09-03
gtk4: setup interface picker + review fixes from #1318

eaescob:
Follow-up to #1318 (the GTK4 + libadwaita interface).

The initial port stubbed the GTK4 **setup screen** down to just the Accept
button and the message log, which dropped the **network-interface selection
entirely** — there was no way to choose which interface to capture on. This
restores the full setup form, rebuilt with libadwaita:

- An `AdwPreferencesGroup` ("Setup") containing:
  - an `AdwComboRow` for the primary (unified) capture interface,
  - an `AdwSwitchRow` "Sniff at startup",
  - an `AdwSwitchRow` "Bridged sniffing" that enables…
  - an `AdwComboRow` for the bridge interface.
- The dropdowns are backed by a `GListModel` of the pcap devices
  (`EcIfaceItem`: name + description). The description is shown; the device
  name is written to `EC_GBL_OPTIONS->iface` / `->iface_bridge`. The primary
  interface is seeded explicitly, since `AdwComboRow` doesn't emit a change
  signal for its initial selection.
- **Accept** switches the engine to bridged sniffing when that toggle is on,
  matching the GTK3 Accept button.

Also restores the setup menu's **"Options"** submenu (Unoffensive, Promisc
mode, Set Netmask), which the GTK3 setup screen exposed and the port dropped.

While here I diffed the GTK4 interface against GTK3 to make sure nothing else
was left behind: **all 48 registered actions and all 38 sniffing-menu items
match**. Only affects `src/interfaces/gtk4/ec_gtk4.c` (GTK3 untouched). Still
5/5 on the `tests/gtk4/` sandbox harness.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1318, opened 2026-09-01
Add GTK4 + libadwaita interface (GTK_BUILD_TYPE=GTK4)

eaescob:
## Summary

Adds a **GTK4 + libadwaita** port of the GTK GUI, selectable with
`-DGTK_BUILD_TYPE=GTK4`. GTK3 stays the default and is left completely
unchanged — the GTK4 interface is a parallel tree under
`src/interfaces/gtk4/`, so this change is purely additive.

## Why now

GTK3 is in maintenance-only mode upstream, and recent GTK / glycin changes
have started to break the GTK3 interface in the field — see #1310, where
`ettercap -G` crashes after idle once glycin restarts its image loaders across
the privilege drop. This gives ettercap a modern, actively-supported toolkit
to move to. The move off GtkTreeView to GtkColumnView may also help with
long-standing rendering glitches some users hit on certain setups (#1231),
though that still needs confirmation from affected users on the new interface.

## What's in it

A full modernization, not a mechanical port:

- **Asynchronous dialogs** everywhere. GTK4 removed `gtk_dialog_run()`; every
  dialog is now build → present → respond, with shared helpers in
  `ec_gtk4_dialogs.c` (`AdwAlertDialog`, `GtkFileDialog`, `AdwToast`).
- **List views** move from `GtkTreeView`/`GtkListStore` to `GtkColumnView`
  over a `GListModel` with a typed `GObject` item per view. This drops the
  hidden pointer column and the fragile `GtkTreeIter`-based deferred name
  resolution.
- **libadwaita widgets**: `AdwApplicationWindow`, `AdwToolbarView` /
  `AdwHeaderBar`, `AdwTabView` for the MDI pages, `AdwAboutDialog`,
  `AdwToast`, and `AdwShortcutsDialog` (with a plain `AdwDialog` fallback for
  libadwaita < 1.8).
- The **keyboard-shortcuts window is generated** from the accelerator table
  instead of the ~1370-line hand-written `ec_gtk3_shortcuts.c`.

Build wiring: a pkg-config-based `FindGTK4.cmake`, `GTK_BUILD_TYPE=GTK4` in
`EttercapLibCheck.cmake`, `HAVE_GTK4`, and a 24.04-only entry in the CI matrix.

## Requirements (for packagers)

- **GTK ≥ 4.12** and **libadwaita ≥ 1.5** (the latter is what provides
  `AdwDialog`/`AdwAlertDialog`). This gates the GTK4 build to **Ubuntu ≥
[...]

PR #1329, opened 2026-09-07
gtk4: address review feedback from #1318 (setup logo, tab-view background, progress dialog)

eaescob:
Addresses @koeppea's review of the GTK4 interface (#1318). These fixes were
meant to ride along with #1320 but landed after that PR was merged, so here
they are as a standalone follow-up.

- **Redundant `toastoverlay` redeclaration** — `ec_gtk4_dialogs.c` re-declared
  it `extern` even though `ec_gtk4.h` already does, producing a
  `-Wredundant-decls` warning. Removed.

- **Setup screen: missing logo / stray editable box** — the setup screen was
  showing the message-log `GtkTextView` (which looks like an editable text
  box) instead of the ettercap logo. GTK3 shows the logo there and keeps the
  message log on the sniffing screen only. Fixed to match: the setup screen
  shows the logo under the setup form; the message log is created only on the
  sniffing screen (`msgbuffer` stays NULL until then, and `gtkui_msg()`
  already no-ops on a NULL buffer).

- **Background logo not replaced when opening a list** — on the sniffing
  screen the logo was a sibling stacked above the tab view, so opening a list
  never replaced it. It is now an overlay on the tab view, shown only while
  the view has no pages (via `notify::n-pages`), so opening a list hides it.

- **`Trying to measure GtkLabel … needs at least N` during scan** — the
  progress dialog set its title as both the `AdwAlertDialog` heading and a
  separate label; the duplicate, unwrapped label caused those warnings. The
  body now holds only the progress bar; the heading carries the title.

Only touches `src/interfaces/gtk4/` (GTK3 untouched). Verified against the
`tests/gtk4/` sandbox harness (5/5). A full GTK3-vs-GTK4 parity audit (menu
actions, context menus, header buttons, accelerators) is in #1320.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)


koeppea:
This PR fixes the observations I reported in #1318 
However I found some other topics, including a segfault, I'll make up a new issue for these observations.

